PR #3, opened 2020-06-26
Configure react with appengine

malachibre:
There's an extra step when you want to deploy with react. You have to deploy the front end by running "yarn build" then "gcloud app deploy" while inside the edith-step directory. Then, you deploy the back-end the regular way using "mvn package appengine:deploy" in the root directory. 

When you want run the app on the test server, it looks like you can only test either the front-end or the back-end at one time. To test the front end run "yarn build" then "yarn local" while in the edith-step directory. And again testing the back-end is the same as usual with "mvn package appengine:run" in the root directory.


dispatch.yaml
```diff
@@ -0,0 +1,9 @@
+# Rules for dispatch file...
+# Put specific cases at the top.
+dispatch:
+# Route the urls that point to the Java backend's API calls
+ - url: "*/api/v1/*"
+   service: backend
+# Route all other urls to the React.js frontend
+ - url: "*/*"
+   service: default
```
Prashantneu8848, line 9: I think we should add a new line at the end of the file ?

edith-step/app.yaml
```diff
@@ -0,0 +1,6 @@
+runtime: nodejs10
+ 
+resources:
+ cpu: 1
+ memory_gb: 0.5
+ disk_size_gb: 11
```
Prashantneu8848, line 6: add a new line ?

dispatch.yaml
malachibre: I added it.

edith-step/app.yaml
malachibre: Added.

malachibre:
> Can everything in the edith-step folder be moved to the root folder?

When I moved everything out the edith-step folder, the app wouldn't compile if the app.js file wasn't in the src. So I moved it and all the related files into the src folder, and it worked fine. But it was inconsistent, because it wasn't inside the webapp folder with index.html. If we're okay with the front-end files not being in the webapp, then we can move it back, but if not we can just keep it all the front-end files separate in the edith-step folder.

PR #4, opened 2020-06-30
ReceiptInput front end form and tests

livseibert5:


edith-step/src/ReceiptInput.js
```diff
@@ -0,0 +1,94 @@
...
+      [e.target.name]: value
+    });
+  }
+
+  render() {
+    return(
+      <div>
+      <h3>Grocery Items</h3>
+      <GroceryList items={this.state.items} />
+      <form onSubmit={this.handleSubmit}>
```
malachibre, line 41: Would it be better to make this form into it's own component?

edith-step/src/ReceiptInput.test.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { shallow, configure, mount } from 'enzyme';
+import ReceiptInput from './ReceiptInput';
+import './setupTests.js'
+
```
malachibre, line 6: Is there a way to specifically test if the GroceryList component was  updated when handleSubmit() is called?

babel.config.js
```diff
@@ -0,0 +1,4 @@
+// babel.config.js
+module.exports = {
+  presets: ['@babel/preset-env', '@babel/preset-react'],
+};
```
Prashantneu8848, line 4: Add a new line at the end of the file ?

edith-step/src/App.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
-import logo from './logo.svg';
-import "./App.css";
+import ReceiptInput from './ReceiptInput';
+//import "./App.css";
```
Prashantneu8848, line 3: Do we not need this css file ?

edith-step/src/ReceiptInput.js
```diff
@@ -0,0 +1,94 @@
+import React, { Component } from 'react';
```
Prashantneu8848, line 1: Do we need to import Component ?

edith-step/src/ReceiptInput.js
```diff
@@ -0,0 +1,94 @@
...
+
+export default class ReceiptInput extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {items: [], text: '', price: 0.0};
```
Prashantneu8848, line 6: can we change the name 'text' to something more specific ? 
is it the receipt name ?

edith-step/src/ReceiptInput.js
```diff
@@ -0,0 +1,94 @@
...
+          <tr key={item.id}>
+            <td>{item.text}</td>
+            <td>{item.price}</td>
+          </tr>
+        ))}
+      </tbody>
+      </table>
+    );
+  }
+}
```
Prashantneu8848, line 94: what do you think of making this a stateless functional component ? It does not seem to have a state.
https://reactjs.org/blog/2015/10/07/react-v0.14.html#stateless-function-components

edith-step/src/ReceiptInput.test.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { render } from '@testing-library/react';
```
Prashantneu8848, line 2: This seems to be not used. Do we need it ?

edith-step/src/ReceiptInput.test.js
```diff
@@ -0,0 +1,55 @@
...
+import { shallow, configure, mount } from 'enzyme';
```
Prashantneu8848, line 3: shallow and configure imports seem to be not used. Do we need it ?

edith-step/src/App.js
livseibert5: It's just the one that displays the spinning logo so we'll need it at some point but it's not tailored to it rn

edith-step/src/ReceiptInput.test.js
```diff
@@ -0,0 +1,55 @@
...
+import ReceiptInput from './ReceiptInput';
+import './setupTests.js'
+
+// Submit button calls handleSubmit when clicked
+it('should call handleSumbit when Submit button is clicked', () => {
+  const handleSubmit = jest.spyOn(ReceiptInput.prototype, 'handleSubmit');
+  const component = mount(<ReceiptInput onSubmit={ handleSubmit }/>);
```
Prashantneu8848, line 10:
```javascript
const component = mount(<ReceiptInput onSubmit={ handleSubmit }/>);
```
and

```javascript
component.unmount();
```
seems to be in every method. Can we use beforeEach() and afterEach() for that ?
https://jestjs.io/docs/en/setup-teardown#repeating-setup-for-many-tests

edith-step/src/ReceiptInput.js
livseibert5: it's the name of the grocery item, i'll change that

edith-step/src/ReceiptInput.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+  component.find('form').simulate('submit');
+  expect(handleSubmit).toBeCalled();
+  component.unmount();
+});
+
+// handleSubmit resets state
```
Prashantneu8848, line 17: niit: add period at the end of the comment.

https://jestjs.io/docs/en/setup-teardown#repeating-setup-for-many-tests

edith-step/src/setupTests.js
```diff
@@ -2,4 +2,7 @@
 // allows you to do things like:
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
-import '@testing-library/jest-dom/extend-expect';
+//import '@testing-library/jest-dom/extend-expect';
```
Prashantneu8848, line 5: Delete the commented import ?

edith-step/src/ReceiptInput.test.js
livseibert5: afterEach() seems to be working well, but beforeEach doesn't work as well due to some tests mocking functions and others not

edith-step/src/ReceiptInput.js
```diff
@@ -0,0 +1,92 @@
...
+export default class ReceiptInput extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {items: [], itemName: '', itemPrice: 0.0};
+    this.handleChange = this.handleChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleSubmit(e) {
+    e.preventDefault();
```
sehtia-raghav, line 12: Is this what checks that the form is set correctly? e.g. that name and price are set? if so, should we add a test for it?

PR #6, opened 2020-07-02
Retrieving user stats

malachibre:
I would appreciate any suggestions on ways to make my testing class a bit more succint.

### DIFF
--- a/src/main/java/com/google/edith/servlets/DataServlet.java
+++ b/src/main/java/com/google/edith/servlets/DataServlet.java
@@ -1,21 +1,64 @@
-
 package com.google.sps.servlets;
 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
 import java.io.IOException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.util.stream.Collectors;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 
 /**
 * Servlet that returns some example content.
 */
-@WebServlet("/api/v1/test-servlet")
+@WebServlet("/api/v1/data-servlet")
 public class DataServlet extends HttpServlet {
 
- @Override
- public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-   response.setContentType("text/html;");
-   response.getWriter().println("Hello world from data servlet!");
- }
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();  
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html;");
+    response.getWriter().println("Hello world from data servlet!");
+
+   
+    updateUserStats();
+  }
+
+
+  public void updateUserStats(String userId) {
+    
+    Query query = new Query("UserStats");
+    PreparedQuery results = datastore.prepare(query);
+    Entity userStats = results.asList(FetchOptions.Builder.withLimit(10))
+                            .stream()
+                            .filter(entity -> ((String) entity.getProperty("userId")).equals(userId))
+                            .collect(Collectors.toList()).get(0);
+
+    List<String> reciepts = (List<String>) userStats.getProperty("Receipts");
+    recipts.add(addRecipt());
+    userStats.setProperty("Receipts", receipts);
+    datastore.put(userStats);
+
+  }
+  
+  public void getUserStats() {
+    Query query = new Query("UserStats");
+    PreparedQuery results = datastore.prepare(query);
+
+  }
+
+  public List<Key> addReceipt() {
+    throw new UnsupportedOperationException();
+  }
+
 }
--- a/src/main/java/com/google/edith/servlets/DataServlet.java
+++ b/src/main/java/com/google/edith/servlets/DataServlet.java
@@ -31,34 +31,5 @@ public class DataServlet extends HttpServlet {
     response.getWriter().println("Hello world from data servlet!");
 
    
-    updateUserStats();
-  }
-
-
-  public void updateUserStats(String userId) {
-    
-    Query query = new Query("UserStats");
-    PreparedQuery results = datastore.prepare(query);
-    Entity userStats = results.asList(FetchOptions.Builder.withLimit(10))
-                            .stream()
-                            .filter(entity -> ((String) entity.getProperty("userId")).equals(userId))
-                            .collect(Collectors.toList()).get(0);
-
-    List<String> reciepts = (List<String>) userStats.getProperty("Receipts");
-    recipts.add(addRecipt());
-    userStats.setProperty("Receipts", receipts);
-    datastore.put(userStats);
-
-  }
-  
-  public void getUserStats() {
-    Query query = new Query("UserStats");
-    PreparedQuery results = datastore.prepare(query);
-
-  }
-
-  public List<Key> addReceipt() {
-    throw new UnsupportedOperationException();
-  }
-
+    }
 }
--- a/src/main/java/com/google/edith/servlets/DataServlet.java
+++ b/src/main/java/com/google/edith/servlets/DataServlet.java
@@ -29,7 +29,5 @@ public class DataServlet extends HttpServlet {
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("text/html;");
     response.getWriter().println("Hello world from data servlet!");
-
-   
     }
 }
--- a/src/main/java/com/google/edith/servlets/Item.java
+++ b/src/main/java/com/google/edith/servlets/Item.java
@@ -1,14 +1,12 @@
-package com.google.sps.servlets;
+package com.google.edith.servlets;
 
  /** Used to hold the properties of an item Entity in datastore. */
 public final class Item {
+  //TODO: Add name, user, category, and receiptId fields.
   double price; 
   long quantity;  
-//   String Name; 
-//   String userId;
   String date; 
-//   String category; 
-  
+ 
   public Item(double price, long quantity, String date) {
     this.price = price;
     this.quantity = quantity;
--- a/src/main/java/com/google/edith/servlets/Item.java
+++ b/src/main/java/com/google/edith/servlets/Item.java
@@ -1,0 +1,17 @@
+package com.google.sps.servlets;
+
+ /** Used to hold the properties of an item Entity in datastore. */
+public final class Item {
+  double price; 
+  long quantity;  
+//   String Name; 
+//   String userId;
+  String date; 
+//   String category; 
+  
+  public Item(double price, long quantity, String date) {
+    this.price = price;
+    this.quantity = quantity;
+    this.date = date;
+  }
+}
--- a/src/main/java/com/google/edith/servlets/Item.java
+++ b/src/main/java/com/google/edith/servlets/Item.java
@@ -2,7 +2,9 @@ package com.google.edith.servlets;
 
  /** Used to hold the properties of an item Entity in datastore. */
 public final class Item {
-  //TODO: Add name, user, category, and receiptId fields.
+  // TODO (malachibre): Add name, user, category, and receiptId fields.
+  // TODO (malachibre): Create a builder for this class inclduing checks 
+  //                    for valid assignments.
   double price; 
   long quantity;  
   String date; 
--- a/src/main/java/com/google/edith/servlets/UserInsights.java
+++ b/src/main/java/com/google/edith/servlets/UserInsights.java
@@ -98,7 +98,7 @@ public final class UserInsights {
     * TODO: Allow for various time periods (only calculates weekly aggregate now).
     * @return A map relating a time period to the spending in that time period.
     */
-  public Map<String, Double> aggregateUserData() { 
+  public Map<String, String> aggregateUserData() { 
     Entity userStats = retreiveUserStats();
     List<Key> items = (List<Key>) userStats.getProperty("Items");
 
@@ -114,8 +114,8 @@ public final class UserInsights {
    * @return Creates a map with keys for each ending day of a weekly period and  
    *         values for the total spending during that period. 
    */
-  public Map<String, Double> calculateWeeklyTotal(List<Key> items)  {
-    Map<String, Double> weeklyTotals = new HashMap<String, Double>();
+  public Map<String, String> calculateWeeklyTotal(List<Key> items)  {
+    Map<String, String> weeklyTotals = new HashMap<String, String>();
     LocalDate currentEndOfWeek;
     try {
       currentEndOfWeek = getEndOfWeek(LocalDate.parse((String) datastore.get(items.get(0))
@@ -132,7 +132,7 @@ public final class UserInsights {
         LocalDate itemDate = LocalDate.parse((String) datastore.get(item).getProperty("date"),
                                              dateFormatter);
         if(ChronoUnit.DAYS.between(itemDate, currentEndOfWeek) < 0) {
-          weeklyTotals.put(currentEndOfWeek.toString(), weeklyTotal);
+          weeklyTotals.put(currentEndOfWeek.toString(), Double.toString(weeklyTotal));
           currentEndOfWeek = getEndOfWeek(itemDate);
           weeklyTotal = 0;
         }
@@ -145,7 +145,7 @@ public final class UserInsights {
       } 
     }
     
-    weeklyTotals.put(currentEndOfWeek.toString(), weeklyTotal);
+    weeklyTotals.put(currentEndOfWeek.toString(), Double.toString(weeklyTotal));
 
     return weeklyTotals;
   }
@@ -169,8 +169,7 @@ public final class UserInsights {
    * @return a Json formatted String of items and an aggregate.
    */
   public String createJson() {
-    Map<String, String> aggregateValues = aggregateUserData().entrySet().stream()
-    .collect(Collectors.toMap(Map.Entry::getKey, entry -> Double.toString(entry.getValue())));
+    Map<String, String> aggregateValues = aggregateUserData();
     String aggregateJson =  gson.toJson(aggregateValues);
     List<Key> itemKeys = (List<Key>) retreiveUserStats().getProperty("Items");
     List<Item> items = itemKeys.stream()
--- a/src/main/java/com/google/edith/servlets/UserInsights.java
+++ b/src/main/java/com/google/edith/servlets/UserInsights.java
@@ -21,7 +21,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Comparator;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -35,8 +35,9 @@ public final class UserInsights {
   
   private String userId;  
   private DatastoreService datastore;
+  private final Gson GSON = new Gson();
   private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd");
-  private static final Gson GSON = new Gson();
+
 
   public UserInsights(String userId) {
     this.userId = userId;
@@ -45,12 +46,13 @@ public final class UserInsights {
 
   /** This should only be called each time a new user makes an accout. */
   public void createUserStats() { 
-     List<Key> items = new ArrayList<>();
+    List<Key> items = new ArrayList<>();
 
-     Entity userStats = new Entity("UserStats");
-     userStats.setProperty("userId", userId);
-     userStats.setProperty("Items", items);
-     datastore.put(userStats);
+    Entity userStats = new Entity("UserStats");
+    userStats.setProperty("userId", userId);
+    userStats.setProperty("Items", items);
+    
+    datastore.put(userStats);
   }
 
   /** 
@@ -59,11 +61,12 @@ public final class UserInsights {
    * @param items - Non-null list of item Keys to be added to the current Item list.
    */
   public void updateUserStats(List<Key> newItems) {
-    Optional<Entity> userStats = retreiveUserStats();
-    if (userStats.isEmpty()) {
+    Optional<Entity> userStatsContainer = retreiveUserStats();
+    if (!userStatsContainer.isPresent()) {
         return;
-    }    
-    List<Key> items = (List<Key>) userStats.get().getProperty("Items");
+    } 
+    Entity userStats = userStatsContainer.get();   
+    List<Key> items = (List<Key>) userStats.getProperty("Items");
     if (items == null) {
       items = newItems;
     } else {
@@ -80,14 +83,15 @@ public final class UserInsights {
    * @return A map relating a time period to the spending in that time period.
    */
   public ImmutableMap<String, String> aggregateUserData() { 
-    Entity userStats = retreiveUserStats().get();
-    if (userStats == null) {
-        Map<String, String> emptyMap = new HashMap<String, String>();
-        emptyMap.put("weeklyAggregate", "");
-        emptyMap.put("items", "");
-        return ImmutableMap.copyOf(emptyMap);
+    Optional<Entity> userStatsContainer = retreiveUserStats();
+    if (!userStatsContainer.isPresent()) {
+     return createDefaultMap(); 
     }
+    Entity userStats = userStatsContainer.get();
     List<Key> items = (List<Key>) userStats.getProperty("Items");
+    if (items == null) {
+      return createDefaultMap();
+    }
 
     /** Assumes dates are in the form yyyy-mm-dd */
     Comparator<Key> SORT_BY_DATE = (Key item1, Key item2) -> {
@@ -98,9 +102,7 @@ public final class UserInsights {
         return 0;
       }
     };
-
     items.sort(SORT_BY_DATE);
-  
     return calculateWeeklyTotal(items);
   }
 
@@ -112,8 +114,16 @@ public final class UserInsights {
   public String createJson() {
     Map<String, String> aggregateValues = aggregateUserData();
     String aggregateJson =  GSON.toJson(aggregateValues);
+    Optional<Entity> userStatsContainer = retreiveUserStats();
+
+    if (!userStatsContainer.isPresent()){
+      return GSON.toJson(createDefaultMap()); 
+    }
+
     List<Key> itemKeys = (List<Key>) retreiveUserStats().get()
                                         .getProperty("Items");
+    // Each item is mapped to an Item object to make their 
+    // properties parseable by GSON.
     List<Item> items = itemKeys.stream()
                          .map(key ->{
                             try {
@@ -121,13 +131,13 @@ public final class UserInsights {
                               return new Item(
                                 (Double) item.getProperty("price"),
                                 (long) item.getProperty("quantity"),
-                                                               (String) item.getProperty("date")
+                                (String) item.getProperty("date")
                               );
                             } catch (EntityNotFoundException e) {
                               return null;
                             }
-                            })
-                            .collect(Collectors.toList());
+                          })
+                          .collect(Collectors.toList());
     String itemsJson = GSON.toJson(items);
     JsonObject userJson = new JsonObject();
     userJson.addProperty("weeklyAggregate", aggregateJson);
@@ -136,10 +146,14 @@ public final class UserInsights {
   }
 
   /**
-   * Finds the UserStats entity corresponding to {@code userId} in datastore.
+   * Finds the UserStats entity corresponding to {@code userId} in datastore
+   * and returns this entity contained inside an Optional.
    * @return UserStats entity for this user
    */
-  private Optional<Entity> retreiveUserStats() {
+  public Optional<Entity> retreiveUserStats() {
+    if (userId == null) {
+      return Optional.empty();
+    }
     Filter idFilter = new FilterPredicate("userId", 
                                 FilterOperator.EQUAL,
                                 userId);
@@ -163,17 +177,16 @@ public final class UserInsights {
    */
   private ImmutableMap<String, String> calculateWeeklyTotal(List<Key> itemKeys)  {
     if (itemKeys == null || itemKeys.isEmpty()) {
-        return ImmutableMap.copyOf(new HashMap<String, String>());
+      return ImmutableMap.copyOf(new LinkedHashMap<String, String>());
     }
-    Map<String, String> weeklyTotals = new HashMap<String, String>();
+    Map<String, String> weeklyTotals = new LinkedHashMap<String, String>();
     LocalDate currentEndOfWeek;
-
     try {
       currentEndOfWeek = getEndOfWeek(LocalDate.parse((String) datastore.get(itemKeys.get(0))
                                           .getProperty("date"), DATE_FORMATTER));
     } catch (EntityNotFoundException e) {
       System.err.println("Error: Entity could not be located");
-      return ImmutableMap.copyOf(new HashMap<String, String>());
+      return ImmutableMap.copyOf(new LinkedHashMap<String, String>());
     }
 
     double weeklyTotal = 0;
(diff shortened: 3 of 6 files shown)


Prashantneu8848:
@malachibre can you please merge master and commit it?

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -158,4 +161,61 @@ public void aggregateData() {
    
     Assert.assertEquals(map, userInsights.aggregateUserData());
   }
+
+//   @Test
```
malachibre, line 165: I am going to refactor this test.

src/main/java/com/google/edith/servlets/DataServlet.java
```diff
@@ -1,21 +1,35 @@
-
 package com.google.sps.servlets;
```
Prashantneu8848, line 1: Should it be changed to ?
```
package com.google.edith.servlets;
```

src/main/java/com/google/edith/servlets/DataServlet.java
```diff
@@ -1,21 +1,35 @@
...
-   response.setContentType("text/html;");
-   response.getWriter().println("Hello world from data servlet!");
- }
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();  
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html;");
+    response.getWriter().println("Hello world from data servlet!");
+
```
Prashantneu8848, line 32: niit: should this space be removed ?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -0,0 +1,17 @@
+package com.google.sps.servlets;
```
Prashantneu8848, line 1: Also, can we change the package name ?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -0,0 +1,17 @@
...
+
+ /** Used to hold the properties of an item Entity in datastore. */
+public final class Item {
+  double price; 
+  long quantity;
```
Prashantneu8848, line 6: Does the quantity have to be long ?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -0,0 +1,17 @@
+package com.google.sps.servlets;
+
+ /** Used to hold the properties of an item Entity in datastore. */
+public final class Item {
+  double price; 
+  long quantity;  
+//   String Name;
```
Prashantneu8848, line 7: Can we remove the commented part ?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -0,0 +1,17 @@
...
+//   String userId;
+  String date; 
+//   String category; 
+  
+  public Item(double price, long quantity, String date) {
+    this.price = price;
+    this.quantity = quantity;
+    this.date = date;
+  }
+}
```
Prashantneu8848, line 17: niit: add a line at the end of the file ?

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,198 @@
+package com.google.sps.servlets;
```
Prashantneu8848, line 1: can we change the package name ?

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,198 @@
...
+      try {
+        return ((String) (datastore.get(item1).getProperty("date")))
+            .compareTo((String) (datastore.get(item2).getProperty("date")));
+      } catch(EntityNotFoundException | NullPointerException e )  {
+        return 0;
+      }
+  };
+  
+  public UserInsights(String userId) {
+    this.userId = userId;
```
Prashantneu8848, line 51: Should we change how the static field is addressed ? I am not sure but something like
'''
 UserInsights.userId = userId;
'''
https://stackoverflow.com/a/25494615/10688048

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,198 @@
...
+    return results.asList(FetchOptions.Builder.withLimit(10)).get(0);
+  }
+
+  /** This should only be called each time a new user makes an accout. */
+  public void createUserStats() { 
+     List<Key> items = new ArrayList<>();
+
+     Entity userStats = new Entity("UserStats");
+     userStats.setProperty("UserId", userId);
+     userStats.setProperty("Items", items);
```
Prashantneu8848, line 73: What do you think of not storing items ? When the user uploads a receipt then both entities UserStates and Item will grow. Instead can we make it consume on the Item Entity ?

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -0,0 +1,221 @@
+package com.google.sps;
```
Prashantneu8848, line 1: Please change the package name.

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -0,0 +1,221 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.sps.servlets.UserInsights;
```
Prashantneu8848, line 14: You might have to change the imports after you change the package name.

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -0,0 +1,221 @@
...
+  public void tearDown() {
+    testHelper.tearDown();
+  }
+
+  @Test
+  public void creatUserInisghts() {
+    // A new UserStats Entity should be added to the datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    userInsights.createUserStats();
+    Assert.assertEquals(1, datastore.prepare(new Query("UserStats")).countEntities());
```
Prashantneu8848, line 54: It seems to be using Deprecated method.
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery
can we use
```
countEntities(FetchOptions fetchOptions)
```
instead

src/test/java/com/google/edith/UserInsightsTest.java
Prashantneu8848: Can you add a TODO for this ?

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,198 @@
...
+    Entity userStats = retreiveUserStats();    
+    List<Key> items = (List<Key>) userStats.getProperty("Items");
+    if (items == null) {
+      items = newItems;
+    } else {
+      items.addAll(newItems);
+    }
+    userStats.setProperty("Items", items);
+    datastore.put(userStats);
+
```
livseibert5, line 92: nit: can delete extra line

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,198 @@
...
+  }  
+  
+  /**
+   * Creates a Json string that contains the weekly aggregate for this user
+   * and the items this user purchased.
+   * @return a Json formatted String of items and an aggregate.
+   */
+  public String createJson() {
+    Map<String, String> aggregateValues = aggregateUserData().entrySet().stream()
+    .collect(Collectors.toMap(Map.Entry::getKey, entry -> Double.toString(entry.getValue())));
```
livseibert5, line 173: I think this part should be indented more

src/test/java/com/google/edith/UserInsightsTest.java
malachibre: I already updated it in the newest commit.

src/main/java/com/google/edith/servlets/DataServlet.java
malachibre: Yea. I changed it.

src/main/java/com/google/edith/servlets/DataServlet.java
malachibre: Done.

src/main/java/com/google/edith/servlets/Item.java
malachibre: I changed it.

src/main/java/com/google/edith/servlets/Item.java
malachibre: Yea. When you retrieve the value from the datastore it's taken as a long, which means I can't convert it to an int.

src/main/java/com/google/edith/servlets/Item.java
malachibre: The reason I have this commented part is because these are theses are the fields we chose in the data model. They just weren't immediately important for this PR, so I left the commented out.

src/main/java/com/google/edith/servlets/Item.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Oh I didn't actually mean to make userId a static field. I removed the static keyword from it.

src/test/java/com/google/edith/UserInsightsTest.java
malachibre: Done.

src/test/java/com/google/edith/UserInsightsTest.java
malachibre: Done.

src/test/java/com/google/edith/UserInsightsTest.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

malachibre: Oh this is from an older commit. I added more in between .stream() and .collect().

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: That's a good point. I'll think about what other fields I can track in the UserStats Entity.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+    datastore.put(userStats);
+  }
+ 
+  /** 
+   * Copmiles the spending using the Item list found in this user's
+   * UserStats Entity in datastore.
+   * TODO: Allow for various time periods (only calculates weekly aggregate now).
+   * @return A map relating a time period to the spending in that time period.
+   */
+  public Map<String, String> aggregateUserData() {
```
sehtia-raghav, line 100: Typically we want to return as specific a type as possible while accepting as generic a type as possible in params.
This might be a great place to start using ImmutableMap as the return type. See
https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableMap.html

You could call ImmutableMap.copyOf. If this is too much overhead you can return a HashMap and pair program the ImmutableMap part later

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /**
+   * Creates a String-Intger map relating weekly periods to spending.
+   * @param items - A list of {@code Key} objects that reference Item entities
+                  - in the datastore.
+   * @return Creates a map with keys for each ending day of a weekly period and  
+   *         values for the total spending during that period. 
+   */
+  public Map<String, String> calculateWeeklyTotal(List<Key> items)  {
```
sehtia-raghav, line 116: Does this method need to be public? Since it's called in aggregate user data which is public?

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+    Map<String, String> weeklyTotals = new HashMap<String, String>();
+    LocalDate currentEndOfWeek;
```
sehtia-raghav, line 118: Let's mention in the javadoc that this method calculates trailing weekly total

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+    try {
+      currentEndOfWeek = getEndOfWeek(LocalDate.parse((String) datastore.get(items.get(0))
+                                          .getProperty("date"), dateFormatter));
+    } catch(EntityNotFoundException e) {
+      System.err.println("Error: Entity could not be located");
+      return null;
```
sehtia-raghav, line 124: Generally we want to avoid nulls in all situations unless absolutely necessary. They tend to be error-prone, difficult to debug, and require more boilerplate for bookeeping. Instead let's return an empty map and the caller can still operate without failing if they don't have a null check

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+   * @param items - A list of {@code Key} objects that reference Item entities
+                  - in the datastore.
+   * @return Creates a map with keys for each ending day of a weekly period and  
+   *         values for the total spending during that period. 
+   */
+  public Map<String, String> calculateWeeklyTotal(List<Key> items)  {
+    Map<String, String> weeklyTotals = new HashMap<String, String>();
+    LocalDate currentEndOfWeek;
+    try {
+      currentEndOfWeek = getEndOfWeek(LocalDate.parse((String) datastore.get(items.get(0))
```
sehtia-raghav, line 120: Could we get a NPE if items is an empty list? Should we add a check?

src/main/java/com/google/edith/servlets/UserInsights.java
sehtia-raghav, line 124: Also, let's say we have a list of 100 items and only one of the items in the list was not found in the datastore. Are we sure we want to return null in that case instead of logging the error and still processing the other 99 items?

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+                                          .getProperty("date"), dateFormatter));
+    } catch(EntityNotFoundException e) {
+      System.err.println("Error: Entity could not be located");
+      return null;
+    }
+    double weeklyTotal = 0;
+    Entity itemEntity;
+    for(Key item : items) {
```
sehtia-raghav, line 128: nit: Let's add a space after keywords like for and if 

`for (Key ...)`

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+      try {
+        itemEntity = datastore.get(item);
```
sehtia-raghav, line 130: Would it help to rename item to "itemKey" and itemEntity to "item"? Just a thought - up to you tho

`item = datastore.get(itemKey)`

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+        LocalDate itemDate = LocalDate.parse((String) datastore.get(item).getProperty("date"),
+                                             dateFormatter);
+        if(ChronoUnit.DAYS.between(itemDate, currentEndOfWeek) < 0) {
+          weeklyTotals.put(currentEndOfWeek.toString(), Double.toString(weeklyTotal));
+          currentEndOfWeek = getEndOfWeek(itemDate);
+          weeklyTotal = 0;
```
sehtia-raghav, line 136: Quickly glancing, I read this as days between itemDate and currentEoW can be negative.. which seems strange. What am I missing about this logic here? Are we assuming that items are sorted by date when passed in?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -0,0 +1,15 @@
+package com.google.edith.servlets;
+
+ /** Used to hold the properties of an item Entity in datastore. */
+public final class Item {
+  //TODO: Add name, user, category, and receiptId fields.
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -0,0 +1,15 @@
...
+  double price; 
+  long quantity;  
+  String date; 
+ 
+  public Item(double price, long quantity, String date) {
+    this.price = price;
+    this.quantity = quantity;
+    this.date = date;
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * This class provides the funttionality to parse specifc user data from
```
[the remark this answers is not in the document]
malachibre: Thanks ahah.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * This class provides the funttionality to parse specifc user data from 
+ * datastore and send an agreggate of that information in a JSON file.
+ */
+public final class UserInsights {
+
+  private String userId;
+
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();  
+  private static final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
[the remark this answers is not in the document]
malachibre: Okay, done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+  private static final Gson gson = new Gson();
+
+  /** Assumes dates are in the form yyyy-MM-dd */
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * This class provides the funttionality to parse specifc user data from 
+ * datastore and send an agreggate of that information in a JSON file.
+ */
+public final class UserInsights {
+
+  private String userId;
+
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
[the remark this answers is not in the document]
malachibre: Okay, I changed it.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();  
+  private static final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+  private static final Gson gson = new Gson();
+
+  /** Assumes dates are in the form yyyy-MM-dd */
+  private static final Comparator<Key> SORT_BY_DATE = (Key item1, Key item2) -> {
+      try {
+        return ((String) (datastore.get(item1).getProperty("date")))
+            .compareTo((String) (datastore.get(item2).getProperty("date")));
+      } catch(EntityNotFoundException | NullPointerException e )  {
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+        return 0;
+      }
+  };
+  
+  public UserInsights(String userId) {
+    this.userId = userId;
```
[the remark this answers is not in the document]
malachibre: I was wondering that too. I could remove it, then make the rest of the methods static. I created this constructor in case we wanted to pass in more information, but so far userId was the only variable I could think of that would be unique for each object.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+    Filter idFilter = new FilterPredicate("UserId", 
+                                FilterOperator.EQUAL,
+                                userId);
+    Query query = new Query("UserStats").setFilter(idFilter);
+    PreparedQuery results = datastore.prepare(query);
+    return results.asList(FetchOptions.Builder.withLimit(10)).get(0);
+  }
+
+  /** This should only be called each time a new user makes an accout. */
+  public void createUserStats() {
```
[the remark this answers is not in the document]
malachibre: Ah okay.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+   * Finds the UserStats entity corresponding to {@code userId} in datastore.
+   * @return UserStats entity for this user
+   */
+  private Entity retreiveUserStats() {
+    Filter idFilter = new FilterPredicate("UserId", 
+                                FilterOperator.EQUAL,
+                                userId);
+    Query query = new Query("UserStats").setFilter(idFilter);
+    PreparedQuery results = datastore.prepare(query);
+    return results.asList(FetchOptions.Builder.withLimit(10)).get(0);
```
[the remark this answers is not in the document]
malachibre: I added it.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+    userStats.setProperty("Items", items);
+    datastore.put(userStats);
+  }
+ 
+  /** 
+   * Copmiles the spending using the Item list found in this user's
+   * UserStats Entity in datastore.
+   * TODO: Allow for various time periods (only calculates weekly aggregate now).
+   * @return A map relating a time period to the spending in that time period.
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,194 @@
...
+      items.addAll(newItems);
+    }
+    userStats.setProperty("Items", items);
+    datastore.put(userStats);
+  }
+ 
+  /** 
+   * Copmiles the spending using the Item list found in this user's
+   * UserStats Entity in datastore.
+   * TODO: Allow for various time periods (only calculates weekly aggregate now).
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: I can do that, I just changed the name of the List to itemKeys.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Yea, it is intended to be negative. If the first parameter is after the second the parameter, Days.between() will be negative. Should I change the parameters so the intended result is that they're positive?

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: No I made it private.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Yea, I added a check.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Yea that makes sense.

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -155,67 +155,69 @@ public void aggregateData() {
     newEntity4.setProperty("date", "2020-07-12");
```
sehtia-raghav, line 155: I think this test is testing two things at once. 1) Two entities with quantity 1 and 2 respectively and (2) entities with more than 1/2 quantities. Ideally, unit tests should be as isolated as possible and test one individual code path at a time (think one assert statement per test in an ideal world). Since this test is kinda large can we split up into two

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -185,10 +190,24 @@ public String createJson() {
...
+   * @return UserStats entity for this user
+   */
+  private Entity retreiveUserStats() {
+    Filter idFilter = new FilterPredicate("UserId", 
+                                FilterOperator.EQUAL,
+                                userId);
+    Query query = new Query("UserStats").setFilter(idFilter);
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> entities = results.asList(FetchOptions.Builder.withLimit(10));
+    return entities.isEmpty() ? null : entities.get(0);
```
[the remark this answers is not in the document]
malachibre: Got it!

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -94,50 +76,73 @@ public void updateUserStats(List<Key> newItems) {
...
+   * Creates a map relating string weekly period keys to string spending values.
+   * This method calculates the trailing based on a date. 
+   * Example: {"06-20-2020" : "53.0"}
    * @param items - A list of {@code Key} objects that reference Item entities
                   - in the datastore.
    * @return Creates a map with keys for each ending day of a weekly period and  
    *         values for the total spending during that period. 
    */
-  public Map<String, String> calculateWeeklyTotal(List<Key> items)  {
+  private ImmutableMap<String, String> calculateWeeklyTotal(List<Key> itemKeys)  {
```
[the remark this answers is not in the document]
malachibre: Okay I added a TODO for this as well.

src/test/java/com/google/edith/UserInsightsTest.java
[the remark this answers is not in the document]
malachibre: Okay I changed them.

malachibre: Okay, I separated it into two parts.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /** 
+   * Updates the Items list in the UserStats Entity in datastore
+   * corresponding to this user.
+   * @param items - Non-null list of item Keys to be added to the current Item list.
+   */
+  public void updateUserStats(List<Key> newItems) {
+    Entity userStats = retreiveUserStats().get();    
+    if (userStats == null) {
```
sehtia-raghav, line 63: I might be mistaken, but I thought this is guaranteed to not be null if we check the optional return from retrieveUserStats().

`Optional<Entity> userStats = retrieveUserStats()`
`if userStats.isEmpty() { ... }`

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,217 @@
...
+    if (itemKeys == null || itemKeys.isEmpty()) {
+        return ImmutableMap.copyOf(new HashMap<String, String>());
+    }
+    Map<String, String> weeklyTotals = new HashMap<String, String>();
+    LocalDate currentEndOfWeek;
+
+    try {
+      currentEndOfWeek = getEndOfWeek(LocalDate.parse((String) datastore.get(itemKeys.get(0))
+                                          .getProperty("date"), DATE_FORMATTER));
+    } catch(EntityNotFoundException e) {
```
sehtia-raghav, line 174: nit: space after 'catch'
`catch (EntityNotFoundException e)`

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    double weeklyTotal = 0;
+    Entity itemEntity;
+    for(Key itemKey : itemKeys) {
+      try {
+        Entity item = datastore.get(itemKey);
+        LocalDate itemDate = LocalDate.parse((String) item.getProperty("date"),
+                                             DATE_FORMATTER);
+        if (ChronoUnit.DAYS.between(itemDate, currentEndOfWeek) < 0) {
```
sehtia-raghav, line 186: Can't find the previous comment, but is there anyway to make this condition more intuitive?

src/main/java/com/google/edith/servlets/UserInsights.java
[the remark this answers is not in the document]
sehtia-raghav: Ah good to know!

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: You're right. I think I removed that part after I committed. I changed it :)

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Yes!

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: You're right, I changed it.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,221 @@
...
+ 
+  /** 
+   * Copmiles the spending using the Item list found in this user's
+   * UserStats Entity in datastore.
+   * TODO (malachibre): Allow for various time periods (only calculates weekly aggregate now).
+   * @return A map relating a time period to the spending in that time period.
+   */
+  public ImmutableMap<String, String> aggregateUserData() { 
+    Entity userStats = retreiveUserStats().get();
+    if (userStats == null) {
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Creates a Json string that contains the weekly aggregate for this user
+   * and the items this user purchased.
+   * @return a Json formatted String of items and an aggregate.
+   */
+  public String createJson() {
+    Map<String, String> aggregateValues = aggregateUserData();
+    String aggregateJson =  GSON.toJson(aggregateValues);
+    List<Key> itemKeys = (List<Key>) retreiveUserStats().get()
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -0,0 +1,221 @@
...
+                              Entity item = datastore.get(key);
+                              return new Item(
+                                (Double) item.getProperty("price"),
+                                (long) item.getProperty("quantity"),
+                                                               (String) item.getProperty("date")
+                              );
+                            } catch (EntityNotFoundException e) {
+                              return null;
+                            }
+                            })
```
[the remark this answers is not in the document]
malachibre: Done.

PR #9, opened 2020-07-08
User stats change

malachibre:
In this Pull Request, the "Items" property in the UserStats Kind was removed. This was done since Items are their own Kind which meant the size of the Items property for all UserStats entities and the total amount of Items entities would increase at the same rate. The benefit of these changes is a reduction in the amount of space being used in datastore. Additionally, the datastore.get(key) operation is not called which is a costly operation. If we chose to go with this implementation over the previous version, then I will will modify the UserInsightsTest.java file to test with these chages.


src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -33,61 +35,64 @@
...
+   */
   public void createUserStats() { 
-     List<Key> items = new ArrayList<>();
-
      Entity userStats = new Entity("UserStats");
-     userStats.setProperty("UserId", userId);
-     userStats.setProperty("Items", items);
+     userStats.setProperty("userId", userId);
+     userStats.setProperty("weekEndDates", new HashSet<String>());
+     userStats.setProperty("weeklyTotals", new ArrayList<String>());
```
malachibre, line 83: I created this as a way to keep track of weekly total spending, but there's no guarantee that the set will be in order, so I'm thinking of creating a list of strings with each element being in the form "weekEndDate = amount".
So one would like "06-20-2020 = 58.75"

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -33,61 +35,64 @@
 
   private String userId;
 
-  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();  
-  private static final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+  private static final DatastoreService datastore = 
+      DatastoreServiceFactory.getDatastoreService();
```
sehtia-raghav, line 39: Any reason to not have this var as a class var that gets initialized via the constructor? This will make testing with a mock/fake datastore service easier.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -33,61 +35,64 @@
 
   private String userId;
 
-  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();  
-  private static final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+  private static final DatastoreService datastore = 
+      DatastoreServiceFactory.getDatastoreService();  
+  private static final DateTimeFormatter dateFormatter =
```
sehtia-raghav, line 40: static final vars should follow upper SNAKE_CASE naming convention.
i.e. `private static final DATE_TIME_FORMATTER = ...`

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -98,54 +103,42 @@ public void updateUserStats(List<Key> newItems) {
...
+    for(Entity item : items) {
+      LocalDate itemDate = LocalDate.parse((String) item.getProperty("date"),
                                              dateFormatter);
-        if(ChronoUnit.DAYS.between(itemDate, currentEndOfWeek) < 0) {
-          weeklyTotals.put(currentEndOfWeek.toString(), Double.toString(weeklyTotal));
-          currentEndOfWeek = getEndOfWeek(itemDate);
-          weeklyTotal = 0;
-        }
+      
+      if(ChronoUnit.DAYS.between(itemDate, currentEndOfWeek) < 0) {
```
sehtia-raghav, line 130: I _think_ this might not be syncd to the latest changes? I feel like we updated this so it reads a little easier (instead of negative time)?

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -33,61 +35,64 @@
...
-  private static final Comparator<Key> SORT_BY_DATE = (Key item1, Key item2) -> {
-      try {
-        return ((String) (datastore.get(item1).getProperty("date")))
-            .compareTo((String) (datastore.get(item2).getProperty("date")));
-      } catch(EntityNotFoundException | NullPointerException e )  {
-        return 0;
-      }
-  };
-  
   public UserInsights(String userId) {
```
sehtia-raghav, line 44: This class seems almost like a 'UserInsightsService" rather than "UserInsights". When I think of "UserInsights", I think of a container class that stores relevant insights info for a particular user, however this class seems to be doing some fetching/processing w.r.t. user insights.

Thinking about this class from a service perspective, it makes a bit more sense to keep the datastore initialized via the constructor and pass the userIds in methods so that this service can be invoked for any user. 

If this class does truly _need_ to be initialized for every user (i.e. create a new UserInsights) that we need insights, I worry there might be a deeper dependency between a "user id" and "insights". And ideally they should be de-coupled or at least be able to be de-coupled.

Wdyt? Apologies for readdressing this, you might just need to refresh my memory. Lmk in case I misunderstood anything

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -11,16 +11,18 @@
 import com.google.appengine.api.datastore.Query.Filter;
```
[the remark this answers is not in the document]
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -33,61 +35,64 @@
...
+   * @return UserStats entity for this user
+   */ 
   private Entity retreiveUserStats() {
-    Filter idFilter = new FilterPredicate("UserId", 
+    Filter idFilter = new FilterPredicate("userId", 
                                 FilterOperator.EQUAL,
                                 userId);
     Query query = new Query("UserStats").setFilter(idFilter);
     PreparedQuery results = datastore.prepare(query);
     return results.asList(FetchOptions.Builder.withLimit(10)).get(0);
```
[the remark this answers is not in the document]
malachibre: Thanks! I was looking for a better way to get the only entity.

src/main/java/com/google/edith/servlets/UserInsights.java
```diff
@@ -98,54 +103,42 @@ public void updateUserStats(List<Key> newItems) {
...
-    }
+    LocalDate currentEndOfWeek = getEndOfWeek(LocalDate.parse((String) 
+                            items.get(0).getProperty("date"), dateFormatter));
     double weeklyTotal = 0;
     Entity itemEntity;
-    for(Key item : items) {
-      try {
-        itemEntity = datastore.get(item);
-        LocalDate itemDate = LocalDate.parse((String) datastore.get(item).getProperty("date"),
+    for(Entity item : items) {
```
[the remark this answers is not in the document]
malachibre: I removed it.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: No, I made it a class variable.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Done.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: Oh no, this is a different version from a different branch. I edited it to how we have it on the master branch.

src/main/java/com/google/edith/servlets/UserInsights.java
malachibre: I understand you. I made a couple to changes that I think defines it more as a service.

PR #12, opened 2020-07-08
Backend for retrieving the cheapest store

livseibert5:
In this pull request:
- ReceiptInput.js and ReceiptInput.test.js: updated to include a quantity field and to send data to deals servlet with axios each time a new item is submitted
- DealItem.java and DealItemTest.java: class that represents items of the same product at different stores so that they can later be compared and the best value item can be returned
-GroceryDataReader.java and GroceryDataReaderTest.java: reads in csv file with the product data, finds the product that corresponds to the user's input and gathers the information on that product from each store. then, it compares that information to return the best deal
-DealsServlet.java and DealsServletTest.java: receives the ReceiptInput data, sends it to GroceryDataReader to be processed, and then sends the best deal back to the frontend

Right now, I have just made the back end for this, and will be developing the front end to display the result in my next PR. I also haven't implemented the Natural Language Processing algorithm yet, so the only input that will get a deal returned is one that is already explicitly listed in the csv.


src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+        DealItem item5 = new DealItem();
+        item5.setStore("Walmart");
+	    item5.setPrice(record[13]);
+	    item5.setWeight(record[14]);
+	    item5.setComment(record[15]);
+        dealItems.add(item5);
+
+        bestItem = getBestDeal(dealItems);
+      }
+	}
```
malachibre, line 69: NIt: Indentation

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+
+        DealItem item5 = new DealItem();
+        item5.setStore("Walmart");
+	    item5.setPrice(record[13]);
+	    item5.setWeight(record[14]);
+	    item5.setComment(record[15]);
+        dealItems.add(item5);
+
+        bestItem = getBestDeal(dealItems);
+      }
```
malachibre, line 68: Nit: Indentation

src/main/java/com/google/edith/servlets/DataServlet.java
```diff
@@ -10,7 +9,7 @@
 /**
 * Servlet that returns some example content.
 */
-@WebServlet("/api/v1/test-servlet")
+@WebServlet("/")
```
malachibre, line 12: Are we sure we want this servlet to link to the home page of the website?

livseibert5: for some reason it was giving me a 503 error unless i linked it to home. not sure why, but i never use the servlet so this can be changed to whatever later on

src/test/java/com/google/edith/DealsServletTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    String json = "{\"itemName\": \"Apple Juice\"}";
+    Mockito.when(request.getReader()).thenReturn(
+        new BufferedReader(new StringReader(json)));
+    
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    Mockito.when(response.getWriter()).thenReturn(writer);
+
+    new DealsServlet().doPost(request, response);
```
Prashantneu8848, line 36: this import seems to be missing ?

edith-step/src/ReceiptInput.js
```diff
@@ -1,28 +1,44 @@
...
     const newItem = {
-      text: this.state.text,
-      price: this.state.price,
+      itemName: this.state.itemName,
+      itemPrice: this.state.itemPrice,
+      itemQuantity: this.state.itemQuantity,
       id: Date.now()
     };
+    
+    const axios = require('axios')
```
Prashantneu8848, line 24: can we use import instead ?
https://stackoverflow.com/a/46677972/10688048

edith-step/src/ReceiptInput.js
```diff
@@ -76,19 +100,18 @@ export default class ReceiptInput extends React.Component {
   }
 }
 
-class GroceryList extends React.Component {
-  render() {
-    return (
-      <table>
+var GroceryList = (props) => {
```
Prashantneu8848, line 103: I was wondering if we can change this var to const ?

edith-step/src/ReceiptInput.test.js
```diff
@@ -4,52 +4,74 @@ import { shallow, configure, mount } from 'enzyme';
 import ReceiptInput from './ReceiptInput';
```
Prashantneu8848, line 4: can we removed the unused imports ? 
render, shallow, configure seem to be not used

src/main/java/com/google/edith/DealItem.java
```diff
@@ -0,0 +1,89 @@
+package com.google.edith;
+
+/** Represents items of the same type from
+  * different stores in order to compare them. */
+public final class DealItem {
+  private String store;
+  private Double price;
```
Prashantneu8848, line 7: Do we really need a Double class wrapper ? What about double primitive ?

src/main/java/com/google/edith/DealItem.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    else if (weight.substring(0, 3).equals("per")) {
+      this.weight = new Double(1);
+    }
+    else if (weight.equals("head")) {
+      this.weight = new Double(1);
+    }
+    else {
+      String num = "";
+      String measure = "";
```
Prashantneu8848, line 45: measure variable does not seem to be used.

src/main/java/com/google/edith/DealItem.java
```diff
@@ -0,0 +1,89 @@
...
+      for (int i=0; i<weight.length(); i++) {
+        if (weight.charAt(i)==' ') {
```
Prashantneu8848, line 47: nit: spacing between operators for consistency.

src/main/java/com/google/edith/servlets/DataServlet.java
Prashantneu8848: can we take this file out of this PR ?
https://git-scm.com/docs/git-checkout#_examples

src/main/java/com/google/edith/servlets/DealsServlet.java
```diff
@@ -0,0 +1,53 @@
+package com.google.edith.servlets;
+
+import com.google.edith.DealItem;
+import com.google.edith.GroceryDataReader;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
```
Prashantneu8848, line 9: InputSream and InputStreamReader does not seem to be used.

livseibert5:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgoogleinterns%2Fstep114-2020%2Fpull%2F12) for more info**.

src/test/java/com/google/edith/DealsServletTest.java
```diff
@@ -0,0 +1,63 @@
+package com.google.edith.servlets;
```
Prashantneu8848, line 1: can we change this to
```
package com.google.edith;
```

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+
+	    DealItem item1 = new DealItem();
+        item1.setStore("Aldi");
+        item1.setPrice(record[1]);
+        item1.setWeight(record[2]);
+        item1.setComment(record[3]);
+        dealItems.add(item1);
+
+        DealItem item2 = new DealItem();
+        item2.setStore("Kroger");
```
sehtia-raghav, line 40: iiuc these store names are pervasive, so we should be able to move them to private static final vars at the top of the class. 
If there are other such common properties that are used in other classes then we can move them to a static utility class as well and make them public static final vars. Lmk if you have any questions 

`private static final String KROGER = "Kroger";`

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+      }
+	}
+		
+	reader.close();
+    return bestItem;
+  }
+
+  /** Gets the $/unit value of each item and
+    * returns the best deal. */
+  private DealItem getBestDeal(List<DealItem> dealItems) {
```
sehtia-raghav, line 77: Although best is pretty straightforward here that we want the cheapest item, since "Best" is somewhat subjective and our definition of best might change, I'd encourage us to rename the method to what best means. getCheapestDeals or getCheapestDealPerItem (again, I'm sure there are better names), but this will clearly indicate when the method is called what it does without having to read internal. Lmk if you feel otherwise or have concerns

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+    double bestVal = 10;
+    DealItem bestItem = dealItems.get(0);
```
sehtia-raghav, line 79: Can we assume dealItems will be non-empty here? If that's an assumption lets add it in the javadoc otherwise let's add an isEmpty check.

src/main/java/com/google/edith/servlets/DealsServlet.java
```diff
@@ -0,0 +1,51 @@
...
+/** Servlet that takes the user input from the receipt
+  * form and uses it to retrieve the best deal. */
+@WebServlet("/receipt-deals")
+public class DealsServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    StringBuilder sb = new StringBuilder();
+    BufferedReader reader = request.getReader();
+    try {
```
sehtia-raghav, line 23: I think you could use a try with resources block here, which will take care of closing any resource for you (in this case the BufferedReader) and should allow you to remove the finally block.

https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/com/google/edith/servlets/DealsServlet.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that takes the user input from the receipt
+  * form and uses it to retrieve the best deal. */
+@WebServlet("/receipt-deals")
+public class DealsServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    StringBuilder sb = new StringBuilder();
```
sehtia-raghav, line 21: Lets use full descriptive names whenever possible

`StringBuilder stringBuilder`

src/test/java/com/google/edith/DealItemTest.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test
+  public void canSetComment() {
+    dealItem.setComment("good deal");
+    Assert.assertEquals("good deal", dealItem.getComment());
+  }
+
+  @Test
+  public void canFindValue() {
```
sehtia-raghav, line 75: Lets update the test name if we update the getValue method name

src/test/java/com/google/edith/DealItemTest.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test
+  public void weightParsedCorrectly() {
+    dealItem.setWeight("15 fl oz");
+    Assert.assertEquals(15.0, dealItem.getWeight(), .01);
+  }
+
+  @Test
+  public void badWeightHandled() {
```
sehtia-raghav, line 56: Similar to how we name methods, it's best to name our test methods as descriptively as possible as well. This makes it easy to debug and isolate errors when there's a breakage or outage. E.g. If I run a presubmit test or check logs and see "negativeWeightReturns0" failed I know exactly what's wrong, whereas I might not understand what "bad" means on sight.

The rule of thumb is that test names should include the condition being tested and the expected outcome. 
E.g. 'multiply_twoNegativeValues_returnsPositiveValue' where multiply is the method, two negative values is the input, and positive value is the output.

Lmk if this makes and if so, let's update test names accordingly. Thanks!

src/test/java/com/google/edith/GroceryDataReaderTest.java
```diff
@@ -0,0 +1,56 @@
...
+      System.err.println();
+    }
+    Assert.assertEquals("Kroger", bestItem.getStore());
+  }
+
+  @Test
+  public void nullReturnedIfNoDeal() {
+    try {
+      bestItem = groceryReader.readFile("bread");
+    } catch (IOException e) {
```
sehtia-raghav, line 41: I think it's okay to let the test throw the exception and the framework will consider it a failed test. This is also standard practice. Lmk in case I missed anything

I.e. 
`public void nullReturnedIfNoDeal() throws Exception`

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+  public DealItem readFile(String itemName) throws IOException {
+    URL csvresource = getClass().getClassLoader().getResource("grocerydata.csv");
+    File groceryDataFile = new File(csvresource.getFile());
+
+    CSVReader reader = new CSVReader(new FileReader(groceryDataFile), ',');
+    DealItem bestItem = null;
+
+    String[] record = null;
+    record = reader.readNext();
+    record = reader.readNext();
```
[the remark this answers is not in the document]
livseibert5: with the file I provide they should never be null, is it still good practice to check first though?

src/main/java/com/google/edith/DealItem.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  public void setWeight(String weight) {
+    if (weight.equals("")) {
+      this.weight = 0.0;
+    }
+    else if (weight.equals("dozen")) {
+      this.weight = 12.0;
+    }
+    else if (weight.substring(0, 3).equals("per")) {
```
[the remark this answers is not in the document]
livseibert5: I checked through the file manually to make sure there weren't any of size 2, but I definitely could've missed one. I can put an else if that sets weight to 0 is the size is 2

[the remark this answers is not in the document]
sehtia-raghav: I think that's fair - in case tomorrow we update the file, ideally existing tests don't break.

src/main/java/com/google/edith/DealItem.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  /**
+    * Grocery items are compared by determining unit
+    * value, which is price divided by weight. This is
+    * because the items at each different store are the
+    * same type, but differ by weight and price. Using the
+    * unit price allows us to compare them more equally.
+    * Since invalid input for weight leads to a default weight
+    * of 0.0, we need to prevent divide by 0 values as well.
```
sehtia-raghav, line 95: Nice! Thanks for adding detail here :)

src/main/java/com/google/edith/GroceryDataReader.java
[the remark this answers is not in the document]
sehtia-raghav: Gotcha in that case a comment should suffice. E.g. "Should never be null with file X"

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,105 @@
...
+
+    CSVReader reader = new CSVReader(new FileReader(groceryDataFile), ',');
+    DealItem cheapestItem = null;
+
+    String[] record = null;
+    record = reader.readNext();
+    record = reader.readNext();
+
+	while ((record = reader.readNext()) != null) {
+      if (record[0].equals(itemName)) {
```
sehtia-raghav, line 39: nit: I think we need to indent the if statement and below here

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,105 @@
...
+        List<DealItem> dealItems = new ArrayList<DealItem>();
+
+        DealItem item1 = new DealItem();
+        item1.setStore(ALDI);
+        item1.setPrice(record[1]);
+        item1.setWeight(record[2]);
+        item1.setComment(record[3]);
+        dealItems.add(item1);
```
sehtia-raghav, line 47: I think if the order of the stores is always the same maybe we could modify lines 42-75 into a loop to avoid duplicate logic
```
private static final ImmutableList<String> STORES = ImmutableList.of(KROGER, TRADERJOES, ..);
for (int i = 0; i < STORES.size(); i++) {
  DealItem item = new DealItem();
  item.setStore(STOERES.get(i));
  item.setPrice(record[i+1]);
  item.setWeight(record[i+2]);
  item.setComment(record[i+3]);
  dealItems.add(item);
}
cheapestItem = getCheapest..();
```
I might have missed something so this may not work, but worth considering if it does.

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,105 @@
...
+
+/** 
+  * Processes data file of product prices and sizes at different
+  * stores and returns the best value item. 
+  */
+public final class GroceryDataReader {
+
+    private static final String ALDI = "Aldi";
+    private static final String KROGER = "Kroger";
+    private static final String TRADERJOES = "Trader Joe's";
```
sehtia-raghav, line 19: nit: For these vars we should use UPPER_SNAKE_CASE so TRADERJOES should be TRADER_JOES. I.e. add a _ where a space should be

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,105 @@
...
+        item5.setPrice(record[13]);
+        item5.setWeight(record[14]);
+        item5.setComment(record[15]);
+        dealItems.add(item5);
+
+        cheapestItem = getCheapestItemPerUnit(dealItems);
+      }
+    }
+		
+	reader.close();
```
sehtia-raghav, line 81: nit: fix indent

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -0,0 +1,105 @@
...
+        item5.setStore(WALMART);
+        item5.setPrice(record[13]);
+        item5.setWeight(record[14]);
+        item5.setComment(record[15]);
+        dealItems.add(item5);
+
+        cheapestItem = getCheapestItemPerUnit(dealItems);
+      }
+    }
+
```
sehtia-raghav, line 80: nit: I think we can remove this extra newline since we have ending punctuation ("}") above

src/test/java/com/google/edith/DealItemTest.java
```diff
@@ -0,0 +1,88 @@
...
+    dealItem = new DealItem();
+  }
+
+  @Test
+  public void instantiate_dealItem_isSuccessful() {
+    Assert.assertTrue(dealItem instanceof DealItem);
+  }
+
+  @Test
+  public void setStore_setsStoreValueToArgument() {
```
sehtia-raghav, line 25: Nice work keeping test cases small/succinct

src/main/java/com/google/edith/GroceryDataReader.java
livseibert5: it works! i had to use record[i*3+1] to get the proper indices

PR #18, opened 2020-07-16
deals data handling

livseibert5:
Changes in this PR:
- DealsServlet.java and DealsServletTest.java: servlet now has a json response so frontend can handle data
- babel.config.js and index.js: configured so i can use async/await in react
- ReceiptInput.js and ReceiptInput.test.js: display deal data on the front end


edith-step/src/ReceiptInput.js
```diff
@@ -9,15 +9,54 @@ export default class ReceiptInput extends React.Component {
...
+      newDeal = {
+        storeName: "no deal found",
+        storePrice: 0
+      }
+    } else {
+      newDeal = {
+        storeName: dealItem.store,
+        storePrice: dealItem.price
+      };
+    }
```
malachibre, line 36: Nit: It's better to use ```===``` instead of ```==``` since ```==``` does not check for data types. Also you might be able to use ternary operator to reduce the amount of lines here.
```
dealItem === "no deal found" ? newDeal = {storeName: "no deal found", storePrice: 0}
                                                :  newDeal = {storeName: dealItem.store, dealItem.price};
```

edith-step/src/ReceiptInput.js
```diff
@@ -9,15 +9,54 @@ export default class ReceiptInput extends React.Component {
...
+
+    return newDeal;
```
malachibre, line 38: Nit: It's better to use ```===``` instead of ```==``` since ```==``` does not check for data types. Also you might be able to use ternary operator to reduce the amount of lines here.
```
dealItem === "no deal found" ? return {storeName: "no deal found", storePrice: 0}
                             : return {storeName: dealItem.store, storePrice: dealItem.price};
```

edith-step/src/ReceiptInput.js
malachibre, line 36: Nit: It's better to use ```===``` instead of ```==``` since ```==``` does not check for data types. Also you might be able to use ternary operator to reduce the amount of lines here.
```
dealItem === "no deal found" ? newDeal = {storeName: "no deal found", storePrice: 0}
                             : newDeal = {storeName: dealItem.store, storePrice: dealItem.price};
```

Prashantneu8848:
Is there any reason to have two different files ```ReceiptInput.js``` ? Also, can we move react components to components folder ?

edith-step/src/components/ReceiptInput.js
```diff
@@ -18,99 +35,148 @@ export default class ReceiptInput extends React.Component {
...
       url: '/receipt-data',
       data: {
         itemName: this.state.itemName,
         itemPrice: this.state.itemPrice,
-        itemQuantity: this.state.itemQuantity
-      }
+        itemQuantity: this.state.itemQuantity,
+      },
     }).then((response) => {
       console.log(response);
```
Prashantneu8848, line 50: Can we remove this debug log ?

edith-step/src/components/ReceiptInput.js
```diff
@@ -18,99 +35,148 @@ export default class ReceiptInput extends React.Component {
...
     const value = e.target.value;
     this.setState({
-      [e.target.name]: value
+      [e.target.name]: value,
     });
   }
 
+  /**
+   * Render grocery list form and items.
+   * @return {html} grocery list form
```
Prashantneu8848, line 74: Is the return type html ?

Prashantneu8848:
Please delete package-lock.json file.

edith-step/src/ReceiptInput.js
corycook:
```javascript
const newDeal = dealItem === "no deal found" ? {storeName: "no deal found", storePrice: 0} : {storeName: dealItem.store, storePrice: dealItem.price};
```

edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,230 @@
...
     this.state = {items: [], itemName: '', itemPrice: 0.0, itemQuantity: 1};
     this.handleChange = this.handleChange.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
   }
 
-  handleSubmit(e) {
+  /**
+   * Sends item info to the receipt-deals servlet and
+   * responds with the cheapest store for that item.
+   * @param {String} name item name
```
sehtia-raghav, line 27: nit: Let's add a space between javadoc and @param

edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,230 @@
...
+    const newDeal = dealItem === 'no deal found' ?
+      {storeName: 'no deal found', storePrice: 0} :
+      {storeName: dealItem.store, storePrice: dealItem.price};
+
+    return newDeal;
+  }
+
+  /**
+   * Send a post request to the receipt-data servlet
+   * with a grocery item in it every time a new item
```
sehtia-raghav, line 53: I'm a bit confused on what it refers to in "with a grocery item in it..", could we clarify here?

edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,230 @@
...
+      url: '/receipt-deals',
+      data: {
+        itemName: name,
+        itemPrice: price,
+        itemQuantity: quantity,
+      },
+    });
+    const dealItem = response.data;
+
+    const newDeal = dealItem === 'no deal found' ?
```
sehtia-raghav, line 44: It would be ideal to have an enum returned for the list of possible stores with one of them being NO_STORE. 
pseudocode:
```
enum {
  TARGET,
  WALMART,
  NO_STORE
}
```

The reason I worry about comparing to 'no deal found' is that if there's accidentally an extra space or something spelled incorrectly then this condition will make it seem like an actual store was found.

The tradeoff is that we would need to update the enum for every new store we add, but for our purposes I think that's ok. Lmk what you think and ofc that can come in a follow-up CL.

edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,230 @@
...
     if (this.state.itemName.length === 0) {
       return;
     }
+
+    const newDeal = await this.getDeal(this.state.itemName,
+        this.state.itemPrice, this.state.itemQuantity);
+    let dealMessage;
+    if (newDeal.storeName == 'no deal found' ||
+        newDeal.storePrice > this.state.itemPrice) {
+      dealMessage = 'no deal found';
```
sehtia-raghav, line 68: FYI if we go with the enum approach, having 'no deal found' as the deal message still makes sense imo

src/main/java/com/google/edith/DealItem.java
```diff
@@ -38,34 +33,26 @@ public void setPrice(String price) {
...
   public void setWeight(String weight) {
     if (weight.isEmpty()) {
       this.weight = 0.0;
-    }
-    else if (weight.equals("dozen")) {
+    } else if (weight.equals("dozen")) {
       this.weight = 12.0;
-    }
-    else if (weight.length() == 2) {
+    } else if (weight.length() == 2) {
```
sehtia-raghav, line 49: we also might be able to combine this check with the line 45 check. it might be a bit strange reading isEmpty and ==2 but if it helps you can leave a comment. Your call just a thought

`if (weight.isEmpty() || weight.length() == 2) {`

edith-step/src/components/ReceiptInput.js
livseibert5: So would this go in the java servlet? Like the response to the post request would use the enum so that the choices we have in the React file are already set

edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,222 @@
...
+    } else {
+      dealMessage = 'Purchase at ' +
+      newDeal.storeName + ' for $' +
+      newDeal.storePrice + '.';
```
corycook, line 74: Use a template literal string:

```javascript
dealMessage = `Purchase at ${newDeal.storeName} for $${newDeal.storePrice}.`;
```

edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,222 @@
...
+    }
```
corycook, line 75: Use a ternary operator and `const`

```javascript
const dealMessage = newDeal.storeName == 'no deal found' || 
    newDeal.storePrice > this.state.itemPrice ? 
  'no deal found' :
  `Purchase at ${newDeal.storeName} for $${newDeal.storePrice}'.`;
```

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -38,16 +38,16 @@ public DealItem readFile(String itemName) throws IOException {
...
       if (record[0].equals(itemName)) {
         List<DealItem> dealItems = new ArrayList<DealItem>();
 
         for (int i = 0; i < STORES.size(); i++) {
           DealItem item = new DealItem();
           item.setStore(STORES.get(i));
-          item.setPrice(record[i*3+1]);
-          item.setWeight(record[i*3+2]);
-          item.setComment(record[i*3+3]);
+          item.setPrice(record[i * 3 + 1]);
```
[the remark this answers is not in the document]
livseibert5: it's 3 because of how the csv file is structured, each store has 3 columns of data so if the loop variable i corresponds to store, the individual cells will be i*3. i'll write a comment for that and store it in a variable

edith-step/src/components/ReceiptInput.test.js
```diff
@@ -1,23 +1,30 @@
...
-import '../setupTests.js'
+import '../setupTests.js';
 
 let component;
 let handleChange;
 let handleSubmit;
 
 beforeEach(() => {
-  handleSubmit = jest.spyOn(ReceiptInput.prototype, 'handleSubmit');
+  handleSubmit = jest.spyOn(ReceiptInput.prototype, 'handleSubmit')
```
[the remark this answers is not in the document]
livseibert5: changing this is making it so that the tests don't register the function as being called, maybe we could go over it on Thursday?

PR #19, opened 2020-07-16
backend for shelf life data

livseibert5:
In this pull request:
- ShelfDataReader.java and ShelfDataReaderTest.java: parse the foodkeeper.json file and return appropriate shelf life data
- doesn't display the data on the front end yet, but can confirm that it works from the tests and what is prints to the command line
- also only works for name input that directly equals the item names in the csv file


edith-step/src/components/ReceiptInput.js
```diff
@@ -1,116 +1,230 @@
...
-      }
-    }).then((response) => {
-      console.log(response);
+        itemQuantity: this.state.itemQuantity,
+      },
     });
 
-    this.setState(state => ({
+    this.setState((state) => ({
       items: state.items.concat(newItem),
```
Prashantneu8848, line 94: what do you think of using ```...``` spread operator for updating arrays intead ?
https://stackoverflow.com/a/26254086/10688048

src/main/java/com/google/edith/servlets/DealsServlet.java
```diff
@@ -1,5 +1,6 @@
 package com.google.edith;
```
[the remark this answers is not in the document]
livseibert5: ideally it would be package com.google.edith.servlets, but if I change it it will give me a 503 error because it thinks theres a servlet w the same name at com.google.edith and com.google.edith.packages. the only way i've found to fix that is to completely delete the project, restart the shell and re clone it. i was thinking maybe i could change it at the very end?

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+   * Retrieves the shelf life data of the specified product. Only looks through pantry and
+   * refrigeration data, as freezing tends to be longer term. Items also tend to have freezing and
+   * fridge data or freezing and pantry data, so removing freezing makes it so that items have only
+   * one set of expiration data.
+   */
+  private String findTime(JsonArray product) {
+    Gson gson = new Gson();
+    String result = "";
+
+    for (int i = 5; i < 27; i++) {
```
malachibre, line 62: Could you add a comment for what the numbers 5 and 27 are? Or create two constants with those values?

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+          potentialMatches.add(product);
+        }
+      }
+    } catch (FileNotFoundException e) {
+      System.out.println(e.getMessage());
+    } catch (IOException e) {
+      System.out.println(e.getMessage());
+    }
+
+    if (potentialMatches.size() == 0) {
```
malachibre, line 45: You could use .isEmpty() instead.

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /** Finds the specified product in the file. */
+  public String readFile(String itemName) {
+    URL jsonresource = getClass().getClassLoader().getResource("foodkeeper.json");
+    File shelfLifeData = new File(jsonresource.getFile());
+
+    JsonParser jsonParser = new JsonParser();
+    List<JsonArray> potentialMatches = new ArrayList<JsonArray>();
+    try (FileReader reader = new FileReader(shelfLifeData)) {
+      JsonObject data = (JsonObject) jsonParser.parse(reader).getAsJsonObject();
```
malachibre, line 26: .parse() is a deprecated method. You can use the static method parseReader() instead.

[](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonParser.html#parseReader(java.io.Reader))

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,98 @@
...
+    String result = "";
+
+    /**
+     * The bounds 6 and 27 correspond to the array indices in the JsonArray that contain shelf life
+     * data. More specifically, the lower bounds marks where data about the type of product stops
+     * and the data about expiration date begins, and the higher bound is where freezing data begins
+     * which, as previously mentioned, is being excluded to minimize duplicate data and very long
+     * term expiration dates.
+     */
+    for (int i = 6; i < 27; i++) {
```
sehtia-raghav, line 70: I think this works perfect for our use case, and also great to also think about other formats or ways we could store this data in a more reliable, strict format. So we don't have to rely on index bounds but can get to the data directly from field names or key values etc. Something like a protocol buffer file would be great, but I'm sure there are others.

Nothing actionable - just brainstorming

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,98 @@
...
+      JsonObject productTimeElement;
+      try {
+        productTimeElement = product.get(i).getAsJsonObject();
+      } catch (IndexOutOfBoundsException e) {
+        break;
+      }
+      String json = gson.toJson(productTimeElement);
+      String[] jsonKeyValuePair = json.split(":");
```
sehtia-raghav, line 78: Can we use Splitter instead and split into a list rather than an Array?
https://guava.dev/releases/23.0/api/docs/com/google/common/base/Splitter.html

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,98 @@
...
+
+      if (jsonKeyValuePair.length == 2) {
+        String expireData = jsonKeyValuePair[1];
+        expireData = expireData.substring(0, expireData.length() - 1);
+        try {
+          Double.parseDouble(expireData);
+          result += expireData + " ";
+        } catch (NumberFormatException e) {
+          String timeUnit = expireData.substring(1, expireData.length() - 1);
```
sehtia-raghav, line 87: I might be mistaken here but is this the same as expireData.substring(1); ?

livseibert5: this line is meant to clip off a trailing quotation mark, expireData.substring(1) would remove the trailing space but not the " at the end

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,98 @@
...
+      JsonObject data = (JsonObject) jsonParser.parseReader(reader).getAsJsonObject();
+      JsonArray sheets = data.getAsJsonArray("sheets");
+      JsonObject productList = sheets.get(2).getAsJsonObject();
+      JsonArray productListData = productList.getAsJsonArray("data");
+
+      for (int i = 0; i < productListData.size(); i++) {
+        JsonArray product = productListData.get(i).getAsJsonArray();
+        JsonObject nameObject = product.get(2).getAsJsonObject();
+        String productName = nameObject.get("Name").getAsString();
+        if (productName.equals(itemName)) {
```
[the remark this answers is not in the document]
livseibert5: I try this in a later PR with the NLP algorithm - stripping the brand data off and only using the common nouns for comparison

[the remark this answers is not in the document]
livseibert5: I could use .contains() instead of .equals() now though and cast everything to lowercase for more matches

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,103 @@
...
+
+    JsonParser jsonParser = new JsonParser();
+    List<JsonArray> potentialMatches = new ArrayList<JsonArray>();
+    try (FileReader reader = new FileReader(shelfLifeData)) {
+      JsonObject data = (JsonObject) jsonParser.parseReader(reader).getAsJsonObject();
+      JsonArray sheets = data.getAsJsonArray("sheets");
+      /*
+       * Get(2) is used because the product data starts at that index, previous data is all
+       * header data on the file.
+       */
```
sehtia-raghav, line 33: super-nit: Typically inline or in-method comments can be done with "//" instead of the javadoc "/**"

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,103 @@
...
+  /** Finds the specified product in the file. */
+  public String readFile(String itemName) {
+    URL jsonResource = getClass().getClassLoader().getResource("foodkeeper.json");
+    File shelfLifeData = new File(jsonResource.getFile());
+
+    JsonParser jsonParser = new JsonParser();
+    List<JsonArray> potentialMatches = new ArrayList<JsonArray>();
+    try (FileReader reader = new FileReader(shelfLifeData)) {
+      JsonObject data = (JsonObject) jsonParser.parseReader(reader).getAsJsonObject();
+      JsonArray sheets = data.getAsJsonArray("sheets");
```
sehtia-raghav, line 29: Is it safe to assume sheets is never empty? Or should we add a check?

src/main/java/com/google/edith/ShelfDataReader.java
sehtia-raghav: Gotcha thanks for clarifying

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+      String json = gson.toJson(productTimeElement);
+      List<String> jsonKeyValuePair = Splitter.on(":").splitToList(gson.toJson(productTimeElement));
+
+      if (jsonKeyValuePair.size() == 2) {
+        String expireData = jsonKeyValuePair.get(1);
+        expireData = expireData.substring(0, expireData.length() - 1);
+        try {
+          Double.parseDouble(expireData);
+          result += expireData + " ";
```
sehtia-raghav, line 90: I'd suggest to use a StringBuilder here as its more performant than string concatenation and typically less error-prone. Wdyt?

https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html

src/main/java/com/google/edith/ShelfDataReader.java
livseibert5: sheets should never be empty with this data file, should I put that in as a comment?

sehtia-raghav: Yep I think that's a valid assumption to add - so the reader doesn't have to go to the actual file

PR #22, opened 2020-07-17
shelf life data front end display

livseibert5:
In this PR:
- ReceiptInput.js and ReceiptInput.test.js: added column to display shelf life data
- DealItem.java and DealItemTest.java: stores expiration data so it can be used on the front end
- GroceryDataReader.java now uses ShelfDataReader.java to incorporate expiration data as well
- ShelfDataReader.java changed to return expiration data differently


edith-step/src/components/ReceiptInput.js
```diff
@@ -15,46 +16,86 @@ export default class ReceiptInput extends React.Component {
...
+
+    const newDeal = await this.getDeal(this.state.itemName,
+        this.state.itemPrice, this.state.itemQuantity);
+
+    const dealMessage = newDeal.storeName === 'NO_STORE' ||
+      newDeal.storePrice > this.state.itemPrice ?
+      'No deal found.' :
+      `Purchase at ${newDeal.storeName} for $${newDeal.storePrice}.`;
+
+    const expirationMessage = newDeal.itemExpiration === 'NO_EXPIRATION' ?
```
Prashantneu8848, line 79: same comment regarding the ternary operator

src/main/java/com/google/edith/servlets/DealsServlet.java
```diff
@@ -33,15 +34,10 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-      response.getWriter().println("no deal found");
-    } else {
-      System.out.println(cheapestItem.getStore());
-      response.getWriter().println(cheapestItem.getStore());
-      response.getWriter().println(cheapestItem.getPrice());
-      response.getWriter().println(cheapestItem.getComment());
-    }
+    Gson gson = new Gson();
+    String responseJson = gson.toJson(cheapestItem);
+    System.out.println(responseJson);
```
Prashantneu8848, line 39: maybe remove this ?

src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -7,4 +7,5 @@
  <sessions-enabled>true</sessions-enabled>
  <runtime>java8</runtime>
  <resource-files><include path="/**.csv" /></resource-files>
+ <resource-files><include path="/**.json" /></resource-files>
```
Prashantneu8848, line 10: should we just use one resource-files tag as it was causing problem while deploying ?

edith-step/src/components/ReceiptInput.js
```diff
@@ -15,46 +16,86 @@ export default class ReceiptInput extends React.Component {
...
     e.preventDefault();
     if (this.state.itemName.length === 0) {
       return;
     }
+
+    const newDeal = await this.getDeal(this.state.itemName,
+        this.state.itemPrice, this.state.itemQuantity);
+
+    const dealMessage = newDeal.storeName === 'NO_STORE' ||
+      newDeal.storePrice > this.state.itemPrice ?
```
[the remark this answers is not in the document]
livseibert5: I had originally had them like this, but the linter flags it and says to put them at the end of the line

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,97 @@
...
+        try {
+          Double.parseDouble(expireData);
+          result += expireData + " ";
+        } catch (NumberFormatException e) {
+          String timeUnit = expireData.substring(1, expireData.length() - 1);
+          if (timeUnit.equals("Days") || timeUnit.equals("Weeks") || timeUnit.equals("Months")) {
+            result += timeUnit + " ";
+          }
+        }
+      }
```
malachibre, line 92: Do you think it's possible to consolidate this into one try-catch block?

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Processes data file of product expiration information to populate future user grocery lists. */
+public class ShelfDataReader {
```
malachibre, line 16: Since this class doesn't have any instance variables, would it be a good idea to make all the methods public to avoid having to instantiate this class each time you need it?

src/main/java/com/google/edith/DealItem.java
```diff
@@ -27,6 +30,7 @@ public String toString() {
   private double weight;
   private String comment;
   private double unitPrice;
+  private String expiration;
```
sehtia-raghav, line 33: "expiration" could be a bit vauge, how about expirationTime? Or something similar

src/main/java/com/google/edith/DealItem.java
```diff
@@ -91,6 +95,44 @@ public void setComment(String comment) {
     this.comment = comment;
   }
 
+  /**
+   * Input string expiration is expected to have the minimum expiration time, the maximum expiration
+   * time, and the unit of time, in no particular order. It parses the string to find the minimum
+   * expiration time and the unit, as that is the data needed to generate new grocery lists based
+   * off of.
+   *
+   * <p>ex: '1.0 2.0 Weeks' -> this.expiration = '1.0 Weeks'
```
sehtia-raghav, line 104: Should the input always expect to include weeks as unit? Or could it be days or months?

src/main/java/com/google/edith/DealItem.java
```diff
@@ -91,6 +95,44 @@ public void setComment(String comment) {
...
+   */
+  public void setExpiration(String expiration) {
```
sehtia-raghav, line 106: Update method name accordingly to var name

src/main/java/com/google/edith/DealItem.java
```diff
@@ -91,6 +95,44 @@ public void setComment(String comment) {
...
+    if (expiration.equals("NO_EXPIRATION")) {
+      this.expiration = expiration;
+      return;
+    }
+
+    String[] expirationPieces = expiration.split(" ");
```
sehtia-raghav, line 112: Can we use splitter here instead and split to a List? 

See https://guava.dev/releases/23.0/api/docs/com/google/common/base/Splitter.html

src/main/java/com/google/edith/DealItem.java
```diff
@@ -91,6 +95,44 @@ public void setComment(String comment) {
...
+    List<Double> range = new ArrayList<Double>();
```
sehtia-raghav, line 113: Can we be a bit more descriptive here for "range". Range of what maybe? or maybe expirationTimes or expirationTimeRange ? Your call

sehtia-raghav, line 113: Could we use an ImmutableList.Builder here instead? 
https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html

I _think_ something like..
```
ImmutableList.Builder<Double> x = ImmutableList.newBuilder()

File: src/main/java/com/google/edith/DealItem.java [hunk 12]
```diff
@@ -91,6 +95,44 @@ public void setComment(String comment) {
...
+
+    Double min = new Double(0);
+    if (range.size() >= 1) {
+      min = range.get(0);
+    }
+    for (Double time : range) {
+      if (time < min) {
+        min = time;
+      }
+    }
```
sehtia-raghav, line 132: Would Collections.min work here?

See https://www.tutorialspoint.com/java/util/collections_min_comparator.htm

src/main/java/com/google/edith/DealItem.java
```diff
@@ -91,6 +95,44 @@ public void setComment(String comment) {
...
+    this.expiration = min + " " + timeMeasurement;
```
sehtia-raghav, line 133: This might be cleaner use a static string that gets formatted.

```
private static final EXPIRATION_TEMPLATE = "%s %s";
this.expiration = String.format(EXPIRATION_FORMAT, min, timeMeasurement);

File: src/main/java/com/google/edith/DealItem.java [hunk 14]
```diff
@@ -121,4 +163,8 @@ public Double getWeight() {
   public String getComment() {
     return comment;
   }
+
+  public String getExpiration() {
+    return expiration;
+  }
```
sehtia-raghav, line 169: Since expiration is a string could we add a javadoc/example for what the caller should expect here? Is it a string expiration date or time until expiration?

src/main/java/com/google/edith/GroceryDataReader.java
```diff
@@ -52,6 +51,11 @@ public DealItem readFile(String itemName) throws IOException {
           item.setPrice(record[storeDataStartColumn + 1]);
           item.setWeight(record[storeDataStartColumn + 2]);
           item.setComment(record[storeDataStartColumn + 3]);
+          if (!expirationTime.isEmpty()) {
+            item.setExpiration(expirationTime);
+          } else {
+            item.setExpiration("NO_EXPIRATION");
+          }
```
sehtia-raghav, line 58: Typically it's best to keep boolean checks positive in the conditional as they're logically easier to follow/read.

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,102 @@
...
+  private static final String NO_EXPIRATION = "NO_EXPIRATION";
+
+  /** Finds the specified product in the file. */
+  public static String readFile(String itemName) {
+    URL jsonresource = ShelfDataReader.class.getClassLoader().getResource("foodkeeper.json");
```
sehtia-raghav, line 22: nit: jsonResource

I think this might need to be syncd with the previous change?

src/main/java/com/google/edith/ShelfDataReader.java
```diff
@@ -0,0 +1,102 @@
...
+      // Get(2) is used because the product data starts at that index, previous data is all
+      // header data on the file.
+      JsonObject productList = sheets.get(2).getAsJsonObject();
+      JsonArray productListData = productList.getAsJsonArray("data");
+
+      for (int i = 0; i < productListData.size(); i++) {
+        JsonArray product = productListData.get(i).getAsJsonArray();
+        JsonObject nameObject = product.get(2).getAsJsonObject();
+        String productName = nameObject.get("Name").getAsString();
+        if (productName.toLowerCase().contains(itemName.toLowerCase())) {
```
sehtia-raghav, line 39: Should we add a comment regarding the upcoming NLP changes. "This algo currently has some false positives, but should be addressed with NLP changes".. along those lines

src/main/java/com/google/edith/DealItem.java
livseibert5: can be days and months too, I'll include that in the javadoc!

PR #28, opened 2020-07-22
connecting receipt input front end to receipt reader back end

livseibert5:
In this PR:
- ReceiptHandler.js and ReceiptHandler.test.js created to merge the deals/expiration feature with the receipt scanner feature. This makes ReceiptInput.js and ReceiptInput.test.js obsolete and they will be deleted. Now the form autopopulates with information from the scanned receipt to allow the user to edit it for more accurate data.
- GroceryNameProcessor.java and GroceryNameProcessorTest.java - ran this nlp algorithm on the csv file to have more generic names, this also runs on every item name input from the form in ReceiptHandler.js to hopefully have more matches and provide better data. I plan to also use this on the json file.
- DealsServlet.java and DealsServletTest.java changed to work better with the new ReceiptHandler.js file


src/main/java/com/google/edith/GroceryNameProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.Document.Type;
+import com.google.cloud.language.v1.EncodingType;
+import com.google.cloud.language.v1.Entity;
+import com.google.cloud.language.v1.EntityMention;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import java.util.ArrayList;
+import java.util.List;
+
+public class GroceryNameProcessor {
```
malachibre, line 14: Nit: Class should be final if it's no other classes are inheriting from it.

pom.xml
```diff
@@ -100,11 +103,34 @@
...
+      <artifactId>mockito-core</artifactId>
+      <version>3.3.3</version>
+      <scope>test</scope>
+    </dependency>
+    <!-->
     <dependency>      
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>29.0-jre</version>
+    </dependency>-->
```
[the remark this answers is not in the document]
livseibert5: my code fails all the tests with the 29.0 jre too

jest.config.js
```diff
@@ -1,9 +1,13 @@
 // jest.config.js
 
 /* eslint max-len: ["error", { "ignoreStrings": true }]*/
+/**
```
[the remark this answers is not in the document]
livseibert5: i dont think so, it's creating a lot of errors

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,286 @@
...
+export default class ReceiptHandler extends React.Component {
+  /**
+   * Constructor
+   * @param {Props} props Setup state.
+   */
+  constructor(props) {
+    super(props);
+    this.state = {userId: '', storeName: '', date: '',
+      name: '', fileUrl: '', totalPrice: 0.0, items: [], deals: []};
+    this.getReceiptData = this.getReceiptData.bind(this);
```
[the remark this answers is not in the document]
livseibert5: what's the advantage of using the arrow function instead of binding?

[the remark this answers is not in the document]
Prashantneu8848: You can read this page 
 https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

pom.xml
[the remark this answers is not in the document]
Prashantneu8848: Hmm, we need to resolve the error because a lot of logic uses  guava classes like ImmutableList, ImmutableMap.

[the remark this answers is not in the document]
livseibert5: i've been using immutable lists without guava using com.google.common.collect.ImmutableList;

edith-step/src/components/ReceiptHandler.js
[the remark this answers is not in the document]
livseibert5: that makes sense, just need to figure out how it works w asynchronous functions as well

src/main/java/com/google/edith/servlets/ReceiptData.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Creates Array of Item objects from list of item description .
+   *
+   * @param user - user who is logged in.
+   * @param extractedData - List of item description.
+   * @return Item[] - array of item objects created from the list of item description.
+   */
+  private Item[] createReceiptItems(User user, List<Map<String, String>> extractedData) {
```
sehtia-raghav, line 74: It's almost always preferred to use Lists over arrays due to immutability, threading, and reliable .equals comparisons and others. Can we return ImmutableList<> here?

src/main/java/com/google/edith/servlets/ReceiptData.java
```diff
@@ -0,0 +1,95 @@
...
+            0.0f,
+            items);
+    return userReceipt;
+  }
+
+  /**
+   * Creates Array of Item objects from list of item description .
+   *
+   * @param user - user who is logged in.
+   * @param extractedData - List of item description.
```
sehtia-raghav, line 71: What is the map inside this list expected to contain? What are the key value pairs?

src/main/java/com/google/edith/servlets/ReceiptFileHandlerServlet.java
```diff
@@ -32,6 +34,10 @@
 @WebServlet("/receipt-file-handler")
 public final class ReceiptFileHandlerServlet extends HttpServlet {
 
+  private BlobKey fileBlobKey;
+  private String expenditureName;
```
sehtia-raghav, line 38: Why declare these here rather than initialize in the method where they are used directly?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -1,17 +1,50 @@
...
+
+  public abstract String category();
+
+  public abstract double price();
+
+  public abstract long quantity();
+
+  public abstract String date();
+
+  public abstract String expiration();
```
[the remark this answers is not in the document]
livseibert5: yes that was in my DealItem.java file though, Item.java was created by Kai and Prashant to use w their features

src/main/java/com/google/edith/servlets/Receipt.java
```diff
@@ -0,0 +1,51 @@
...
+
+/** Encapsulate User info and logout url. */
+public final class Receipt {
+  private final String userId;
+  private final String storeName;
+  private final String date;
+  private final String name;
+  private final String fileUrl;
+  private final float totalPrice;
+  private final Item[] items;
```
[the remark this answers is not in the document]
livseibert5: this is also a file I didn't create, I don't want to change it and mess up Kai and Prashant's files that depend on it

src/main/java/com/google/edith/servlets/ReceiptData.java
livseibert5: also someone else's file, just an unmerged dependency

src/main/java/com/google/edith/servlets/ReceiptFileHandlerServlet.java
livseibert5: same as above, this file is an unmerged dependency

src/main/java/com/google/edith/servlets/ReceiptData.java
livseibert5: also an unmerged file from someone else's PR, I think the data model is set up with Item as an array

src/main/java/com/google/edith/GroceryNameProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.cloud.language.v1.EncodingType;
+import com.google.cloud.language.v1.Entity;
+import com.google.cloud.language.v1.EntityMention;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import java.util.ArrayList;
+import java.util.List;
+
+public final class GroceryNameProcessor {
+  public String process(String text) throws Exception {
+    try (LanguageServiceClient language = LanguageServiceClient.create()) {
```
corycook, line 16: This line is causing the test failure. This should be mocked.

```java
public final class GroceryNameProcessor {
  
  private final LanguageServiceClient client;

  public GroceryNameProcessor() {
    this.client = LanguageServiceClient.create();
  }

  GroceryNameProcessor(LanguageServiceClient client) {
    this.client = client;
  }
}
```

If you need to create these on a per-request basis you can use a Provider

```java

public final class GroceryNameProcessor {

  public interface LanguageServiceClientProvider {
    LanguageServiceClient get(); // possibly add "throws Exception" if necessary
  }
  
  private final LanguageServiceClientProvider clientProvider;

  public GroceryNameProcessor() {
    this.clientProvider = () -> LanguageServiceClient.create();
  }

  GroceryNameProcessor(LanguageServiceClientProvider clientProvider) {
    this.clientProvider = clientProvider;
  }

  public String process(String text) throws Exception {
    try (LanguageServiceClient language = clientProvider.get()) {
      ...
    }
  }
}
```

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+     * dates.
+     *
+     * @param {Event} e change event
+     */
+    this.handleSubmit = async (e) => {
+      e.preventDefault();
+      let price = 0;
+      this.state.items.forEach((item) => {
+        price += item.price * item.quantity;
+      });
```
corycook, line 69: Use `reduce` for this.

```javascript
const price = this.state.items.reduce((sum, item) => sum + item.price * item.quantity, 0);
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+          name: this.state.name,
+          fileUrl: this.state.fileUrl,
+          totalPrice: price,
+          items: this.state.items,
+        },
+      });
+
+      const data = response.data;
+      this.setState((state) => ({
+        deals: data,
```
corycook, line 87: `response.data` can probably be directly referenced from here without issue.

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+          date: this.state.date,
+          name: this.state.name,
+          fileUrl: this.state.fileUrl,
+          totalPrice: price,
+          items: this.state.items,
+        },
+      });
+
+      const data = response.data;
+      this.setState((state) => ({
```
corycook, line 86: Previous state isn't used to compute next state so shorthand can be used

```javascript
this.setState({deals: response.data});
```

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+          storeName: this.state.storeName,
+          date: this.state.date,
+          name: this.state.name,
+          fileUrl: this.state.fileUrl,
+          totalPrice: price,
+          items: this.state.items,
+          deals: this.state.deals,
+        },
+      });
+      axios({
```
corycook, line 116: Should `await` be used here?

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+          itemPrice: this.state.itemPrice,
+          itemQuantity: this.state.itemQuantity,
+          itemDate: this.getDate(),
+          itemReceiptId: this.state.itemReceiptId,
+          itemCategory: this.state.itemCategory,
+        },
+      }).catch((err) => {
+        console.log(err);
+      });
+      this.setState({});
```
corycook, line 131: I don't think this will change the state. Is this to force a re-render of the component?

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+      deals: [],
+    }));
+  }
+
+  /**
+   * Handles changes to the name of a grocery item.
+   * @param {number} i index of item
+   * @param {Event} e change event
+   */
+  handleNameChange(i, e) {
```
corycook, line 161: avoid single character function arguments (except in small arrow functions where they are obvious from context or unimportant)

edith-step/src/components/ReceiptHandler.js
```diff
@@ -0,0 +1,307 @@
...
+                    name="name"
+                    value={item.name}
+                    onChange={this.handleNameChange.bind(this, i)}/>
+                </div>
+                <div className="col-lg-2">
+                  <input type="number"
+                    className="price form-control"
+                    name="price"
+                    value={item.price}
+                    onChange={this.handlePriceChange.bind(this, i)}/>
```
corycook, line 250: use arrow functions instead of `bind`

```jsx
onChange={element => this.handlePriceChange(index, element)}
```

PR #31, opened 2020-07-23
Update javalint.yaml

Prashantneu8848:
Change it to make an automatic commit on PR


Prashantneu8848:
Auto commit GitHub action will not be used.

PR #33, opened 2020-07-24
fixed serviceWorker.js so it would pass the linter

livseibert5:
service worker kept failing the linting tests so i changed it so our main branch can pass the tests

### DIFF
--- a/edith-step/src/serviceWorker.js
+++ b/edith-step/src/serviceWorker.js
@@ -11,15 +11,19 @@
 // opt-in, read https://bit.ly/CRA-PWA
 
 const isLocalhost = Boolean(
-  window.location.hostname === 'localhost' ||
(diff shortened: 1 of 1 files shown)


edith-step/src/serviceWorker.js
```diff
@@ -54,88 +61,102 @@ export function register(config) {
...
-            window.location.reload();
+      .then((response) => {
+        /**
+         * Ensure service worker exists, and that we really are getting
+         * a JS file.
+         */
+        const contentType = response.headers.get('content-type');
+        if (
+          response.status === 404 ||
+      (contentType != null && contentType.indexOf('javascript') === -1)
```
sehtia-raghav, line 131: nit: this indent seems a bit strange, but might just be me

PR #34, opened 2020-07-27
queries grocery list items based on expiration date

livseibert5:
In this pull request:
- GroceryList.js and GroceryList.test.js retrieves expired item data from the backend and displays it
- App.js added a button to show the GroceryList component so it doesn't conflict with the ReceiptHandler component
- GroceryListServlet.java and GroceryListServletTest.java gets list of most recent receipts and sends them to QueryItems to be searched for expired items
- QueryItems.java and QueryItemsTest.java this file finds the time difference in days between the current date and the date the receipt was stored. it then converts the item's expiration date to days, and if the days passed is greater than the item's expiration it sends them to the servlet to be sent to the front end to be displayed on the shopping list
- ShelfDataReader.java and ShelfDataReaderTest.java shortened the expiration date finder algorithm and made it more accurate
- Items.java and ItemsTest.java overrode equals and hashcode so that using a hashset would get rid of duplicate items
- GroceryNameProcessor.java and GroceryNameProcessorTest.java significantly more sophisticated testing
- DealsServlet.java and DealsServletTest.java

Future goals:
- improve layout/find a better place to put the data


edith-step/src/App.js
```diff
@@ -1,19 +1,83 @@
...
+};
+
+/** Main webpage for the website. */
 class App extends Component {
+  /** Constructor */
+  constructor() {
+    super();
+    this.state = {'chartType': LineChart, 'dateFilter': '',
+      'categoryFilter': '', showGroceryList: true};
+    this.handleGroceryListShow = this.handleGroceryListShow.bind(this);
```
Prashantneu8848, line 23: let's use arrow function here instead of using bind.

pom.xml
```diff
@@ -100,11 +103,29 @@
...
+      <scope>test</scope>
+    </dependency>
+    <!--
     <dependency>      
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>29.0-jre</version>
+    </dependency>-->
+    <dependency>
+      <groupId>org.mockito</groupId>
```
Prashantneu8848, line 120: are these two mockito dependencies different ? They have the same groupId

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Finds the items on past receipts that have expired by the time of the user's next shopping trip.
+ */
+public class QueryItems {
```
sehtia-raghav, line 25: Consider making this class final

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+    this.userService = userService;
+  }
+
+  public String findExpiredItems() {
+    List<Receipt> receipts = queryReceipts();
+    Set<Item> itemsToBuy = new HashSet<Item>();
+
+    for (Receipt receipt : receipts) {
+      double daysPassedSinceReceipt = findTimePassed(receipt.getDate());
+      Item[] items = receipt.getItems();
```
sehtia-raghav, line 41: Let's prefer Lists over object arrays

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+
+      for (Item item : items) {
+        String expiration = item.expiration();
+        if (expiration.equals("NO_EXPIRATION")) {
+          continue;
+        }
+
+        List<String> expirationNumberAndUnit = Splitter.on(" ").splitToList(expiration);
```
sehtia-raghav, line 49: I think we can move the " " into a private static final str at the top of the file

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+        String unit = "";
+
+        for (String expirationPiece : expirationNumberAndUnit) {
+          try {
+            number = (double) Double.parseDouble(expirationPiece);
+          } catch (NumberFormatException e) {
+            unit = expirationPiece;
+          }
+        }
+        if (unit.toLowerCase().equals("weeks")) {
```
sehtia-raghav, line 60: Let's also move "weeks" and "months" to private static final vars

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+          number = number * 7;
+        } else if (unit.toLowerCase().equals("months")) {
+          number = number * 30;
+        }
+
+        if (number < daysPassedSinceReceipt) {
```
sehtia-raghav, line 66: nit: daysPassedSinceReceiptExpired? or just daysPassedSinceExpiry?

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+
+    Gson gson = new Gson();
+    return gson.toJson(itemsToBuy);
+  }
+
+  /**
+   * Determines the number of days between when the receipt was stored in datastore and the current
+   * time.
+   */
+  private double findTimePassed(String receiptDateString) {
```
sehtia-raghav, line 80: Can we be a bit more descriptive with the name as I had to read the javadoc/code to find out. Time passed since when? Maybe findTimePassedSinceStored or findStoredDuration or findReceiptStorageDuration. Totally open since these are meh

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+    Date currentDate = new Date();
+    Date receiptDate;
+    try {
+      receiptDate = new SimpleDateFormat("yyyy-mm-dd").parse(receiptDateString);
+    } catch (ParseException e) {
+      receiptDate = new Date();
+    }
+    long timePassedSinceReceipt = currentDate.getTime() - receiptDate.getTime();
+    double daysPassedSinceReceipt = (timePassedSinceReceipt / (1000 * 60 * 60 * 24)) / 24;
```
sehtia-raghav, line 89: Want to leave a small comment for the division math here?

src/main/java/com/google/edith/QueryItems.java
sehtia-raghav: Also might even say findDurationDays or append Days to indicate the method returns days as units always

sehtia-raghav, line 80: Since we are returning days, any reason to not return an int?

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+      String userId = (String) receiptEntity.getProperty("userId");
+      String storeName = (String) receiptEntity.getProperty("storeName");
+      String date = (String) receiptEntity.getProperty("date");
+      String name = (String) receiptEntity.getProperty("name");
+      String fileUrl = (String) receiptEntity.getProperty("fileUrl");
+      float totalPrice = (float) ((double) receiptEntity.getProperty("price"));
+      Receipt receipt = new Receipt(userId, storeName, date, name, fileUrl, totalPrice, items);
+      receipts.add(receipt);
+    }
+    return receipts;
```
sehtia-raghav, line 117: Let's return ImmutableList. You could use an ImmutableList.Builder or use ImmutableList.copyOf()

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /**
+   * Creates an array of Item objects from entites of kind Item found in the datastore.
+   *
+   * @param entities - entities of kind Item found in datastore.
+   * @return Item[] - array of Item objects.
+   */
+  public Item[] createItemObjects(List<Entity> entities) {
```
sehtia-raghav, line 126: We should always prefer Lists (ideally ImmutableLists) to object arrays

src/main/java/com/google/edith/QueryItems.java
```diff
@@ -0,0 +1,151 @@
...
+    PreparedQuery receiptResults = datastore.prepare(query);
+    List<Entity> receiptEntities = receiptResults.asList(FetchOptions.Builder.withLimit(3));
+    List<Receipt> receipts = new ArrayList<>();
+
+    for (Entity receiptEntity : receiptEntities) {
+      Key entityKey = receiptEntity.getKey();
+      Query itemQuery = new Query("Item", entityKey);
+      PreparedQuery results = datastore.prepare(itemQuery);
+      List<Entity> itemEntities = results.asList(FetchOptions.Builder.withLimit(Integer.MAX_VALUE));
+      Item[] items = createItemObjects(itemEntities);
```
sehtia-raghav, line 107: We should get a List<Item> here

src/main/java/com/google/edith/servlets/GroceryListServlet.java
```diff
@@ -0,0 +1,38 @@
...
+public class GroceryListServlet extends HttpServlet {
+  private UserService userService;
+  private DatastoreService datastoreService;
+  private QueryItems queryItems;
+
+  public GroceryListServlet(
+      UserService userService, DatastoreService datastoreService, QueryItems queryItems) {
+    this.userService = userService;
+    this.datastoreService = datastoreService;
+    this.queryItems = queryItems;
```
sehtia-raghav, line 23: Do we need the QueryItems as a separate params if they are built using the userService and datastoreService? Or does this help in testing?

src/main/java/com/google/edith/servlets/GroceryListServlet.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/grocery-list-query")
+public class GroceryListServlet extends HttpServlet {
+  private UserService userService;
+  private DatastoreService datastoreService;
+  private QueryItems queryItems;
```
sehtia-raghav, line 17: Can we make these private and final?

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -48,4 +48,18 @@ public static Builder builder() {
...
+    Item itemObject = (Item) item;
+    if (itemObject.name().equals(this.name()) && itemObject.userId().equals(this.userId())) {
+      return true;
+    }
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    return this.userId().hashCode() * 11;
```
sehtia-raghav, line 63: Just verifying we have tests for these right?

sehtia-raghav, line 63: Why *11?

src/test/java/com/google/edith/QueryItemsTest.java
```diff
@@ -0,0 +1,124 @@
...
+          .setEnvIsLoggedIn(true)
+          .setEnvAuthDomain("gmail")
+          .setEnvIsAdmin(true)
+          .setEnvEmail("user@gmail.com");
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+    testHelper.setUp();
+    query = new QueryItems(datastore, userService);
```
sehtia-raghav, line 43: Can query be directly initialized in line 22?

src/main/java/com/google/edith/QueryItems.java
livseibert5: the issue is I need to mock this class to test another class and i get a can't mock final class error - is there a way around that?

src/main/java/com/google/edith/QueryItems.java
livseibert5: it's days passed since the receipt was stored, i'll change it to that

src/main/java/com/google/edith/QueryItems.java
livseibert5: the issue here and with the comment below is that the data model takes an array. i could have the function return an immutable list but would have to turn it back into an array to go into datastore

src/main/java/com/google/edith/servlets/GroceryListServlet.java
livseibert5: yes this constructor is used for testing purposes only, otherwise the default constructor is used

src/main/java/com/google/edith/servlets/Item.java
livseibert5: just submitted tests

src/main/java/com/google/edith/QueryItems.java
livseibert5: i think im going to change it to use a java function to convert, i've been having some issues w it

src/main/java/com/google/edith/servlets/Item.java
livseibert5: just changed it to include the name of the item instead of a random int

edith-step/src/components/GroceryList.js
```diff
@@ -0,0 +1,62 @@
...
+
+  /** Gets expired items from grocery-list-query servlet. */
+  async getItemData() {
+    const response = await axios({
+      method: 'get',
+      url: '/grocery-list-query',
+      responseType: 'json',
+    });
+    const itemsList = response.data;
+    this.setState({items: itemsList});
```
corycook, line 31:
```javascript
componentDidMount() {
  this.isMounted = true;
}

componentWillUnmount() {
  this.isMounted = false;
}

...

if (this.isMounted) {
  this.setState({items: itemsList});
}
```

edith-step/src/components/GroceryList.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+it('mounts properly', () => {
+  expect(component.exists()).toBe(true);
+});
+
+it('calls getItemData on mount', () => {
+  expect(getItemData).toBeCalled();
+});
+
+it('displays items', () => {
```
corycook, line 26: Make this test async.

edith-step/src/components/GroceryList.test.js
```diff
@@ -0,0 +1,40 @@
...
+  const newItem = {
+    name: '',
+    price: 0.0,
+    quantity: 1,
+  };
+  component.setState({items: newItem});
+  expect(component.state('items')).toBe(newItem);
+
+  const promise = new Promise(getItemData);
+  promise.then(() => {
```
corycook, line 37: await this promise, it might cause the test to fail since getItemData might not trigger this promise. I'm not sure about the internals of the spy.

edith-step/src/components/ReceiptHandler.js
```diff
@@ -190,13 +186,45 @@ export default class ReceiptHandler extends React.Component {
...
+  addItem(e) {
+    const newItem = {
+      userId: this.state.userId,
+      name: '',
+      price: 0.0,
+      quantity: 0,
+      category: '',
+      expireDate: '',
+    };
+    this.setState({items: this.state.items.concat(newItem)});
```
corycook, line 203: `[...this.state.items, newItem]`

edith-step/src/components/ReceiptHandler.js
```diff
@@ -190,13 +186,45 @@ export default class ReceiptHandler extends React.Component {
...
+  }
+
+  /**
+   * Calculates the current date in yyyy-mm-dd format
+   * @return {string} The current date
+   */
+  getDate() {
+    const date = new Date(Date.now());
+    let month = date.getMonth() + 1;
+    month < 10 ? month = '0' + month.toString() : month = month.toString();
```
corycook, line 213: `month = month < 10 ? `0${month}` : month.toString();`

edith-step/src/components/ReceiptHandler.js
```diff
@@ -190,13 +186,45 @@ export default class ReceiptHandler extends React.Component {
...
+    this.setState({items: this.state.items.concat(newItem)});
+  }
+
+  /**
+   * Calculates the current date in yyyy-mm-dd format
+   * @return {string} The current date
+   */
+  getDate() {
+    const date = new Date(Date.now());
+    let month = date.getMonth() + 1;
```
corycook, line 212: try to use `const`

src/main/java/com/google/edith/QueryItems.java
corycook: You could create an interface and mock that (+ implement it here)

src/main/java/com/google/edith/servlets/GroceryListServlet.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/grocery-list-query")
+public class GroceryListServlet extends HttpServlet {
+  private final UserService userService;
+  private final DatastoreService datastoreService;
+  private final QueryItems queryItems;
+
+  public GroceryListServlet(
```
corycook, line 19: Remove 'public' since this is just for testing

src/main/java/com/google/edith/servlets/Item.java
```diff
@@ -48,4 +48,18 @@ public static Builder builder() {
...
+      return true;
+    }
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    return this.userId().hashCode() * this.name().hashCode();
+  }
 }
```
corycook, line 65: AutoValue would implement `equals`, `hashCode`, and `toString` for you

src/test/java/com/google/edith/QueryItemsTest.java
```diff
@@ -0,0 +1,123 @@
...
+        createReceiptEntity("12345", "Whole Foods", "2020-08-01", "sunday", "url2", 3.5f);
+    createItemEntity(receipt1, "12345", "Apple Juice", 5, 5.6f, "fruit", "6.0 Days");
+    createItemEntity(receipt1, "12345", "Peanut Butter", 6, 4.8f, "protein", "1.0 Weeks");
+    createItemEntity(receipt2, "12345", "Peanut Butter", 5, 2.5f, "protein", "1.0 Weeks");
+    createItemEntity(receipt3, "12345", "Bread", 3, 4.6f, "bread", "2.0 Weeks");
+  }
+
+  // Creates a Receipt Entity and stores it in Datastore.
+  private Entity createReceiptEntity(
+      String userId, String storeName, String date, String name, String fileUrl, float price) {
```
corycook, line 93: Methods with 3+ arguments are generally discouraged, especially native type arguments. A fix would be to use a container class for these values (e.g. AutoValue) with a Builder that has named setters.

```java
createReceiptEntity(Receipt.newBuilder().setUserId(...).setStoreName(...) [...] .build());
```

PR #35, opened 2020-07-27
Use Stream API

Prashantneu8848:
Use java stream API.

### DIFF
--- a/src/test/java/com/google/edith/UserInsightsTest.java
+++ b/src/test/java/com/google/edith/UserInsightsTest.java
@@ -231,7 +231,7 @@ public final class UserInsightsTest {
   }
 
   /**
-   * Creates a list of Item keys
+   * Creates a list of Item keys.
    *
    * @param start the number of the first element in the list
    * @param count the number of the elements after first element
--- a/src/test/java/com/google/edith/UserInsightsTest.java
+++ b/src/test/java/com/google/edith/UserInsightsTest.java
@@ -9,11 +9,13 @@ import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
 import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -231,16 +233,14 @@ public void createJson_withUnkownUser_returnsDefaultJson() {
...
    */
-  private List<Key> createTestKeyList(int startIndex, int endIndex) {
-    List<Key> items = new ArrayList<>();
-    for (int i = startIndex; i < endIndex; i++) {
-      items.add(KeyFactory.createKey("Item", "Item" + i));
-    }
-    return items;
+  private List<Key> createTestKeyList(int start, int count) {
+    return IntStream.range(start, start + count)
+        .mapToObj(i -> KeyFactory.createKey("Item", "Item" + i))
```
malachibre, line 242: Nit: You could probably change the name from 'i' to 'index', but it's not that confusing with the way it is now.

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -231,16 +233,14 @@ public void createJson_withUnkownUser_returnsDefaultJson() {
...
+   * @param count the number of the elements after first element
    * @return a list of keys with a number assigned to each of them "item0, item1... itemN"
    */
-  private List<Key> createTestKeyList(int startIndex, int endIndex) {
-    List<Key> items = new ArrayList<>();
-    for (int i = startIndex; i < endIndex; i++) {
-      items.add(KeyFactory.createKey("Item", "Item" + i));
-    }
-    return items;
+  private List<Key> createTestKeyList(int start, int count) {
```
sehtia-raghav, line 240: It's best to return as specific a type that we can. Also we want to prefer immutable objects wherever possible. So let's return ImmutableList<Key> here

src/test/java/com/google/edith/UserInsightsTest.java
```diff
@@ -231,16 +233,14 @@ public void createJson_withUnkownUser_returnsDefaultJson() {
   /**
    * Creates a list of Item keys
```
sehtia-raghav, line 234: nit: period at end of statement

PR #36, opened 2020-07-28
Search entities

Prashantneu8848:
Proividing a modal box for user to put filers which will search the datastore using those filters and shows the results.
Files changed brief:
SearchModalBox.js -> Modal box to provide user interface to to select their searching criteria.
SearchModalBox.test.js -> Tests for SearchModalBox.js
SearchResult.js -> UI to present the found results after searching the datastore.
SearchResult.test.js -> Tests for SearchResult.js
TopNavbar.js -> Added a section for search feature
SearchServlet.java -> Search which gets fiters from the user and provides the search result.
SearchServiceImpljava -> Prepares query based on the filters and returns entities from datastore matching the filters
SearchServiceImplTest -> Tests for SearchServiceImpl class.
SearchServletTest.java -> Tests for SearchServlet class
SearchService.java -> Interface which provides basic operation for searching in datastore.


src/test/java/com/google/edith/SearchServletTest.java
```diff
@@ -0,0 +1,241 @@
...
+        Item.builder()
+            .setUserId("12345")
+            .setName("apple")
+            .setPrice(1.5)
+            .setQuantity(2)
+            .setDate("date")
+            .setCategory("fruit")
+            .setExpiration("expireDate")
+            .build();
+    // Item item1 = new Item("12345", "apple", 1.5f, 2, "fruit", "date");
```
livseibert5, line 208: nit: can this be deleted since the builder is being used now

src/test/java/com/google/edith/SearchServletTest.java
```diff
@@ -0,0 +1,241 @@
...
+        Item.builder()
+            .setUserId("12345")
+            .setName("apple")
+            .setPrice(1.5)
+            .setQuantity(2)
+            .setDate("date1")
+            .setCategory("fruit")
+            .setExpiration("expire1")
+            .build();
+    // Item item1 = new Item("12345", "apple", 1.5f, 2, "fruit", "date");
```
livseibert5, line 235: nit: same here

livseibert5 (approved):
LGTM

src/test/java/com/google/edith/SearchServletTest.java
Prashantneu8848: Deleted it.

src/test/java/com/google/edith/SearchServletTest.java
Prashantneu8848: Deleted it.

src/main/java/com/google/edith/services/SearchServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+import com.google.edith.servlets.Item;
+import com.google.edith.servlets.Receipt;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Service to create query and handle searching on entities. */
+public final class SearchServiceImpl implements SearchService {
+
+  private final DatastoreService datastore;
+  private UserService userService;
```
malachibre, line 41: Is there a reason userService isn't final?

Prashantneu8848: It should have been final. Thanks for catching that.

src/main/java/com/google/edith/services/SearchServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Creates a filter either on name and date.
+   *
+   * @param name -name property of the entity.
+   * @param date - date property of the entity.
+   * @return Filter -filter options to filer on entites.
+   */
+  private Filter prepareFilter(String name, String date) {
+    String loggedInUserId = userService.getCurrentUser().getUserId();
+    List<Filter> filters = new ArrayList<>();
```
malachibre, line 142: I'm not sure if it's necessary, but you could use an ImmutableList here too, and just put

malachibre (approved):

src/main/java/com/google/edith/services/SearchServiceImpl.java
Prashantneu8848: I am building a filter here with each filter obtained from the user. I am not quite sure it we can make ImmutableList and add to the list as we go ? lmk if I am missing something.

src/main/java/com/google/edith/interfaces/SearchService.java
```diff
@@ -0,0 +1,54 @@
...
+// limitations under the License.
+
+package com.google.edith.interfaces;
+
+import com.google.appengine.api.datastore.Entity;
+import com.google.common.collect.ImmutableList;
+import com.google.edith.servlets.Item;
+import com.google.edith.servlets.Receipt;
+
+/** Operations for searching entities in Datastore with given filters. * */
```
sehtia-raghav, line 22: nit: extra javadoc at the end

src/main/java/com/google/edith/interfaces/SearchService.java
```diff
@@ -0,0 +1,54 @@
...
+public interface SearchService {
+  /**
+   * Creates an immutable list of Receipt objects from entites of kind Receipt found in the
```
sehtia-raghav, line 25: nit; "ImmutableList"

sehtia-raghav, line 25: nit: this javadoc is a little verbose. Hb ".. list of Receipt objects from receipt entities"

src/main/java/com/google/edith/servlets/SearchServlet.java
```diff
@@ -0,0 +1,94 @@
...
+    String date = getParameter(request, "date").orElse("");
+    String sortOrder = getParameter(request, "sort-order").orElse("");
+    String sortOnProperty = getParameter(request, "sort-on").orElse("").toLowerCase();
+
+    ImmutableList<Entity> entities =
+        searchService.findEntityFromDatastore(name, date, kind, sortOrder, sortOnProperty);
+
+    for (Entity entity : entities) {
+      System.out.println(entity);
+    }
```
sehtia-raghav, line 74: remove print statements

src/main/java/com/google/edith/servlets/SearchServlet.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Returns Optional of String from form in FE.
+   *
+   * @param request - id of the user who is logged in.
+   * @param name - name of the field.
+   * @return Optional<String> -
+   */
+  private Optional<String> getParameter(HttpServletRequest request, String name) {
+    return Optional.ofNullable(request.getParameter(name));
+  }
```
sehtia-raghav, line 93: Not sure if we need a helper method for this one line, but if you feel it helps readability feel free to keep it

src/main/java/com/google/edith/services/SearchServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  public ImmutableList<Entity> findEntityFromDatastore(
+      String name, String date, String kind, String sortOrder, String sortOnProperty) {
+    Query query = prepareQuery(name, date, kind, sortOrder, sortOnProperty);
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> entities = results.asList(FetchOptions.Builder.withLimit(Integer.MAX_VALUE));
+    return ImmutableList.copyOf(entities);
+  }
```
sehtia-raghav, line 102: This method is a bit unwieldy with so many params and it would be ideal to break up the method into each param so the caller can directly use findEntityWithName() rather than using this method and giving a bunch of empty strings. You can add a todo if thats easier :)

src/test/java/com/google/edith/SearchServiceImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+          .setEnvIsLoggedIn(true)
+          .setEnvAuthDomain("gmail")
+          .setEnvIsAdmin(true)
+          .setEnvEmail("user@gmail.com");
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+    testHelper.setUp();
+    searchService = new SearchServiceImpl(datastore, userService);
```
sehtia-raghav, line 61: Can this be initialized on line 41 itself? you'd have to move after line 43, but seems reasonable? or do we need a new instantiation every test case

src/test/java/com/google/edith/SearchServletTest.java
```diff
@@ -0,0 +1,239 @@
...
+        Item.builder()
+            .setUserId("12345")
+            .setName("apple")
+            .setPrice(1.5)
+            .setQuantity(2)
+            .setDate("date")
+            .setCategory("fruit")
+            .setExpiration("expireDate")
+            .build();
+    Item[] items = {item};
```
sehtia-raghav, line 208: Ideally we should prefer Lists to object arrays wherever possible

sehtia-raghav (approved):

src/test/java/com/google/edith/SearchServletTest.java
Prashantneu8848: The Receipt class has Items field which is an array.

src/main/java/com/google/edith/services/SearchServiceImpl.java
Prashantneu8848: Thanks for the suggestion. The problem with that approach might be that we just do not want to return entities matching the name only, we have to build the query with all the filters provided and then return it. And also if we are querying each time, then the total number of times we are querying the datastore will increase which is what I wanted to avoid.

src/main/java/com/google/edith/servlets/SearchServlet.java
Prashantneu8848: It is to handle null when the request parameter of that kind is not in the request

src/test/java/com/google/edith/SearchServiceImplTest.java
Prashantneu8848: Thanks for the suggestion. I have initialized it in only one place.

src/main/java/com/google/edith/servlets/SearchServlet.java
Prashantneu8848: Changes made.

src/main/java/com/google/edith/interfaces/SearchService.java
Prashantneu8848: Thanks for the suggestion. Changes made.

Prashantneu8848: Changes made.

src/main/java/com/google/edith/interfaces/SearchService.java
Prashantneu8848: Changes made.

src/main/java/com/google/edith/interfaces/SearchService.java
```diff
@@ -0,0 +1,53 @@
...
+   *
+   * @param name -name property of the entity.
+   * @param date - date property of the entity.
+   * @param kind - kind of the entity stored in datastore.
+   * @param sortOrder - order to sort the entities.
+   * @param sortOnProperty - property on which to sort the order.
+   * @return ImmutableList<Entity> - ImmutableList of entites found from the query.
+   */
+  ImmutableList<Entity> findEntityFromDatastore(
+      String name, String date, String kind, String sortOrder, String sortOnProperty);
```
corycook, line 52: What kind of entities is this method retrieving? The return type should be a class obejct, like Item.

src/main/java/com/google/edith/services/SearchServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+              (float) ((double) entity.getProperty("price")),
+              items.toArray(new Item[items.size()]));
+
+      receipts.add(receipt);
+    }
+    return ImmutableList.copyOf(receipts);
+  }
+
+  @Override
+  public ImmutableList<Item> createItemObjects(ImmutableList<Entity> entities) {
```
corycook, line 75: This method can be static, and probably private and called from another method.

src/main/java/com/google/edith/services/SearchServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+   * @param name -name property of the entity.
+   * @param date - date property of the entity.
+   * @return Filter -filter options to filer on entites.
+   */
+  private Filter prepareFilter(String name, String date) {
+    String loggedInUserId = userService.getCurrentUser().getUserId();
+    List<Filter> filters = new ArrayList<>();
+    filters.add(new FilterPredicate("userId", FilterOperator.EQUAL, loggedInUserId));
+
+    if (!name.isEmpty()) {
```
corycook, line 144: These should probably be Optional<String> instead of expecting an empty string.

src/main/java/com/google/edith/servlets/SearchServlet.java
```diff
@@ -0,0 +1,90 @@
...
+    String sortOnProperty = getParameter(request, "sort-on").orElse("").toLowerCase();
+
+    ImmutableList<Entity> entities =
+        searchService.findEntityFromDatastore(name, date, kind, sortOrder, sortOnProperty);
+
+    if (kind.equals("Receipt")) {
+      receipts = searchService.createReceiptObjects(entities);
+    } else {
+      items = searchService.createItemObjects(entities);
+    }
```
corycook, line 76: This method should really send the values directly rather than setting `receipts` and `items`. These variables are not specific to a user and this implementation will behave unusually when multiple people are using the application.

corycook (approved):
LGTM, some suggestions but feel free just to acknowledge them.

src/main/java/com/google/edith/interfaces/SearchService.java
Prashantneu8848: it was just a general method in an interface to retrieve entities of certain kind given filters on name or date.

src/main/java/com/google/edith/services/SearchServiceImpl.java
Prashantneu8848: I am calling it from a different method. The query might be to retrieve Items only so we might still have to expose this method to servlet in that case.

src/main/java/com/google/edith/services/SearchServiceImpl.java
Prashantneu8848: That's a great idea. I am using optional to retrieve those parameters from the request but converting it to empty string if null.
Acknowledged.

src/main/java/com/google/edith/servlets/SearchServlet.java
Prashantneu8848: That's a great idea. I will incorporate that change in next PR.

PR #40, opened 2020-08-05
Add user auth

malachibre:
Change in UserStatsServlet.java: userId is now specific to each user

### DIFF
--- a/src/main/java/com/google/edith/servlets/UserStatsServlet.java
+++ b/src/main/java/com/google/edith/servlets/UserStatsServlet.java
@@ -35,11 +35,12 @@ public class UserStatsServlet extends HttpServlet {
   public UserStatsServlet() {
     this.datastore = DatastoreServiceFactory.getDatastoreService();
     this.userInsights = new UserInsightsService();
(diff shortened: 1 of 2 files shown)


corycook:
Fix java lint error.

src/main/java/com/google/edith/servlets/UserStatsServlet.java
```diff
@@ -26,21 +29,23 @@
 
   private final DatastoreService datastore;
   private final UserInsightsInterface userInsights;
+  private final String userId;
 
   public UserStatsServlet() {
     this.datastore = DatastoreServiceFactory.getDatastoreService();
     this.userInsights = new UserInsightsService();
+    this.userId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
```
sehtia-raghav, line 37: This might be a bit less brittle if the UserService is a class variable/ dependency instead. This would give the caller control over what dependency to pass and make testing easier

sehtia-raghav: You can still set the userId as a class variable from the UserService var. Lmk if I missed anything

